PR #30, opened 2017-01-04
[RFC] Clean up install.sh

mhinz:
I used shellcheck on it and fixed all immediate issues.

Two questions:

1. You often saved `$?` in `ret`, but didn't use `ret` anywhere yet. I guess it's just not implemented yet?
2. In two places you do `file "$HOME/.vim" | grep "$HOME/.SpaceVim"`. What is that supposed to do? I can't see how the `$HOME/.SpaceVim` should ever be contained in the output of `$HOME/.vim`. Maybe I'm missing something here.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -9,62 +9,61 @@ Red='\033[0;31m'          # Red
 Blue='\033[0;34m'         # Blue
 
 need_cmd () {
-    if ! command -v "$1" > /dev/null 2>&1;then
-        printf "${Red}need '$1' (command not found)${Color_off}\n"
-        exit 0
+    if ! hash "$1" &>/dev/null; then
+        echo -e "${Red}need '$1' (command not found)${Color_off}"
+        exit 1
     fi
 }
 
 fetch_repo () {
-    if [ -d "${HOME}/.SpaceVim" ];then
-        cd ${HOME}/.SpaceVim
-        git pull
-        ret="$?"
-        printf "${Blue}Successfully update SpaceVim${Color_off}\n"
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -9,62 +9,61 @@ Red='\033[0;31m'          # Red
 Blue='\033[0;34m'         # Blue
 
 need_cmd () {
-    if ! command -v "$1" > /dev/null 2>&1;then
-        printf "${Red}need '$1' (command not found)${Color_off}\n"
-        exit 0
+    if ! hash "$1" &>/dev/null; then
+        echo "${Red}need '$1' (command not found)${Color_off}"
```
wsdjeg: 2017-01-05_1362x555

@mhinz thanks for PR, I think we should use `printf` to make colored output.

mhinz: Ah, I forgot the `-e` for interpreting the escape sequences. Fixed.

wsdjeg: LGTM.

mhinz:
What about those two questions above, though? :-) The first one can be ignored for now.. but I don't understand 2.

wsdjeg:
@mhinz  2 is just for make sure ~/.vim is a symbolic of ~/.SpaceVim
2017-01-05_630x132

mhinz:
Oh, the `file` from OSX doesn't work that way:

```
$ echo foo > file
$ ln -s file link
$ /usr/bin/file link
link: ASCII text
```
I'll fix that.

wsdjeg:
Oh, thanks, sorry I do not know this about OSX.

PR #59, opened 2017-01-07
Enable https for spacevim.org

### LINKED ISSUE #25: Any Network Attackers Can Execute Arbitrary Code to Destroy Users' System during Installation
>  Oh, yeah! Good old curl | sh via HTTP. Fucking awesome. I wonder how many people get unknowingly pwned while executing such curl | sh instructions. It's not hard at all to MitM HTTP, detect if shellcode is being transmitted, then add your own instructions to it. Or you can just compile a list of known sources that encourage to execute curl | sh and MitM only them if you want to make your exploits even more discreet. What a nice way to get yourself your own little botnet. Way to go! There's a catch though. You need access to infrastructure of ISP. But is it that hard to get such access? No.

# Short-term Solution

Using `https://spacevim.github.io` or another HTTPS URL. If you are going to use a service like CloudFlare, make sure the upstream server also has HTTPS, and CloudFlare SSL/TLS is set to "Strict" mode.

# Long-term Solution

Creating a proper installation process.

wsdjeg:
close #25

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ For learning about Vim in general, read
 ##### Install
 
 ```sh
-curl -sLf https://raw.githubusercontent.com/SpaceVim/SpaceVim/dev/install.sh | bash
+curl -sLf https://spacevim.org/install.sh | bash
 ```
 before use SpaceVim, you should install the plugin by `call dein#install()`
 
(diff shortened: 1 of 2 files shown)


catkins:
This seems to have upset cloudflare

screen shot 2017-01-07 at 8 10 41 pm

wsdjeg:
@catkins just fix it , can you try again?

catkins:
Looks like it's working again 👍

PR #65, opened 2017-01-07
Silent checkinstall on startup

wsdjeg:
Issue : https://github.com/Shougo/dein.vim/issues/177


wsdjeg:
@tracyone I will try to fix startup checkinstall

PR #67, opened 2017-01-07
Add support for Go with defaults

phacops:
Add support for Go with defaults (from https://github.com/spf13/spf13-vim).


autoload/SpaceVim/plugins.vim
```diff
@@ -191,6 +191,7 @@ let s:plugins.lang = [
             \ ['isundil/vim-irssi-syntax',               { 'on_ft' : 'irssi'}],
             \ ['lervag/vimtex',                          { 'on_ft' : 'tex'}],
             \ ['vimperator/vimperator.vim',              { 'on_ft' : 'vimperator'}],
+            \ ['fatih/vim-go',                           { 'on_ft' : 'go'}],
```
wsdjeg, line 194: need loadconf  or loadconf_before if the config file need to be loaded before vim-go.

wsdjeg (approved):

autoload/SpaceVim/plugins.vim
wsdjeg: merged local with some fix.

PR #77, opened 2017-01-09
Add support for custom dir

### LINKED ISSUE #71: Split user-config and officia-config
Make a opinion, take a look at the entire directory structure,
`~ /. SpaceVim` link to ` ~ /. Vim`,no problem, but for user-defined configuration, no corresponding to centrally manage user configuration, such as **spacemacs**, it has one `. spacemacs.d` directory, default generates a `init. el`, as the user configuration entry, users can customize here configuration, and can be incorporated into their VCS, implementation and the separation of user layer, or so for user-defined configuration is very difficult to maintain. My this spacemac.d independent configuration, for example, `https://github.com/co1inlee/dotfile/tree/master/spacemacs.d`, full control of their own, do not rely on and spacemacs. Although the README says, there is a `.local.vim` configuration file, but just a file, may be somewhat limited

wsdjeg:
I have add this feature, but need more test. close #71

### DIFF
--- a/autoload/SpaceVim.vim
+++ b/autoload/SpaceVim.vim
@@ -140,8 +140,15 @@ function! SpaceVim#loadCustomConfig() abort
     if filereadable(custom_glob_conf)
         exe 'source ' . custom_glob_conf
     endif
+    if isdirectory(expand('~/.SpaceVim.d/'))
+        set runtimepath^=~/.SpaceVim.d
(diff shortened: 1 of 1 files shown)


wsdjeg:
I have test it, it works well for me.

colinlabs:
thanks, i will update!

PR #98, opened 2017-01-12
Update readme for NERD Commenter

### LINKED ISSUE #97: Suggestion: Change tinycomment for tcomment or smth else
This is just a suggestion to change `tinycomment` for something else. `tcomment` comes to mind since it's still maintained and tinycomment has seen it's last commit 2 years ago.

btw, `<leader>-v` for commenting is also not working out of the box

wsdjeg:
Fix #97

### DIFF
--- a/config/plugins/vimfiler.vim
+++ b/config/plugins/vimfiler.vim
@@ -57,6 +57,7 @@ function! s:vimfilerinit()
     silent! nunmap <buffer> gr
     silent! nunmap <buffer> gf
     silent! nunmap <buffer> -
+    silent! nunmap <buffer> s
 
     nnoremap <silent><buffer> gr  :<C-u>Denite grep:<C-R>=<SID>selected()<CR> -buffer-name=grep<CR>
     nnoremap <silent><buffer> gf  :<C-u>Denite file_rec:<C-R>=<SID>selected()<CR><CR>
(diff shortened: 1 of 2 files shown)


cserb (approved):
Everything works for me except `<leader>+cs` in that it destroys the formatting. But this looks more like a nerd commenter problem together with ruby

wsdjeg:
Open issue in it's repo maybe better, and I think should provide the diff  before comment and after it. 

so can we merge this PR?

config/plugins/vimfiler.vim
```diff
@@ -57,6 +57,7 @@ function! s:vimfilerinit()
     silent! nunmap <buffer> gr
     silent! nunmap <buffer> gf
     silent! nunmap <buffer> -
+    silent! nunmap <buffer> s
```
cserb, line 60: not sure what this line has to do with nerd commenter

cserb (approved):
yes the problem described has nothing to with this PR and was just as a fyi (maybe wrong place)

we are good to go

config/plugins/vimfiler.vim
wsdjeg: Because I use s as my windows func leader, so I need disable it in vimfiler buffer.

cserb: 👍

wsdjeg: thanks, merged.

PR #113, opened 2017-01-15
Add lang/go layer

### LINKED ISSUE #108: go support
- [x] autocompletion

- [x] syntax check

- [x] go to definetion

- [x] debug

wsdjeg:
close #108

### DIFF
--- a/autoload/SpaceVim/layers/lang/go.vim
+++ b/autoload/SpaceVim/layers/lang/go.vim
@@ -1,9 +1,9 @@
 function! SpaceVim#layers#lang#go#plugins() abort
+    let plugins = [['fatih/vim-go', { 'on_ft' : 'go', 'loadconf_before' : 1}]]
     if has('nvim')
-        return ['zchee/deoplete-go', {'on_ft' : 'go', 'build': 'make'}]
-    else
-        return ['fatih/vim-go', { 'on_ft' : 'go', 'loadconf_before' : 1}]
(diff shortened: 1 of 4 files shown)


autoload/SpaceVim/layers/lang/go.vim
```diff
@@ -0,0 +1,12 @@
+function! SpaceVim#layers#lang#go#plugins() abort
+    if has('nvim')
+        return ['zchee/deoplete-go', {'on_ft' : 'go', 'build': 'make'}]
+    else
+        return ['fatih/vim-go', { 'on_ft' : 'go', 'loadconf_before' : 1}]
```
phacops: Why not load `vim-go` no matter what? As far as I know it's neovim compatible (support for neovim is in beta but still) and it has more than just completion but running `go fmt` on save or `go vet`/`go imports`.

wsdjeg: there is a projrect nvim-go, and it is remote plugin, I will look into it.

wsdjeg: @phacops I have enable vim-go for neovim, but in neovim, deoplete is the default completion tools , so I use deoplete-go source, and ignore omni func from vim-go.
@zchee I want to use nvim-go for neovim though, if it is stabe please let me know, thanks for your work on nvim-go.

PR #119, opened 2017-01-17
Fix #118

### LINKED ISSUE #118: Go support is not loaded anymore
### Expected behavior, english is recommend
When opening a Go file, I expect to be able to run a command from the vim-go plugin (for example GoImports). Better, it would run automatically (as the default config is configured that way).

### Environment Information
- OS: Arch Linux
- neovim version: 0.1.7

### The reproduce ways from Vim starting (Required!)
Open a file ending by `.go`, type `<leader>GoImports` and it fails.

### Output of the ":message" command, and ":echo SpaceVim#logger#viewLog()"
E492: Not an editor command: GoImports
[ SpaceVim ] : logger file /home/phacops/.SpaceVim/.SpaceVim.log does not exists!

phacops:
Fix go layer and moved the default configuration.


config/plugins_before/vim-go.vim
```diff
@@ -1,20 +0,0 @@
...
-    au FileType go nmap <Leader>s <Plug>(go-implements)
-    au FileType go nmap <Leader>i <Plug>(go-info)
-    au FileType go nmap <Leader>e <Plug>(go-rename)
-    au FileType go nmap <leader>r <Plug>(go-run)
-    au FileType go nmap <leader>b <Plug>(go-build)
-    au FileType go nmap <leader>t <Plug>(go-test)
-    au FileType go nmap <Leader>gd <Plug>(go-doc)
-    au FileType go nmap <Leader>gv <Plug>(go-doc-vertical)
-    au FileType go nmap <leader>co <Plug>(go-coverage)
-augroup END
```
wsdjeg, line 20: Why move this to config? Config is for current layer not only vim-go.

phacops: I noticed the config wasn't loaded when the layer was loaded. Having the config it here loads it properly.

wsdjeg: ok, merged, Thanks for the PR. and I will try to fix config loading problem.

PR #131, opened 2017-01-18
Fixed to work with vim-plug

camdencheek:
Docs claim to support vim-plug, but I could not get it to work
by default. I fixed zvim#plug#add() and s:install_manager to
at least work minimally.

Still unfixed is that using vim-plug does not seem to respect
lazy-loading rules.


wsdjeg:
thanks , I think we should parser the argv of the plugin, by default I use dein` options dict.

wsdjeg:
S: parser need to be improved. PR welcome, or I will fixed latter.

PR #147, opened 2017-01-21
add support for puppet

igalic:
The "standard" puppet vim plugin has moved to Voxpupuli. This plugin
requires tabular and syntastic, both of which we already provide.


wsdjeg:
Thanks, merged, the lint error is not for this PR. I will fix it latter.

igalic:
> manifests/init.pp|105 col 7 error| [autoloader_layout] php not in autoload module layout

i will have to add a basic configuration to disable this check…
but also the tabular plugin doesn't seem to be interacting properly…

wsdjeg:
pr welcome, I do not family with pp file.

PR #151, opened 2017-01-22
Add a download guard

Leonidas-from-XIV:
To prevent the shell from executing incomplete downloads it might be useful to wrap the whole install script in a guard function instead of executing whatever part was downloaded.


wsdjeg:
Thanks!

wsdjeg:
@Leonidas-from-XIV we need change last line `guard` to `guard $@`, this will make install script support argv.

Leonidas-from-XIV:
Great. But as a side-note, the CLA bot seems to be a bit misconfigured, it wants me to sign a CLA for SAP?

wsdjeg:
Yeah, it just is a bot. Request for cla.

PR #158, opened 2017-01-23
Fix logger

wsdjeg:


autoload/SpaceVim/options.vim
```diff
@@ -0,0 +1,7 @@
+function! SpaceVim#options#list() abort
+    let list = []
+    for var in getcompletion('g:spacevim_','var')
```
wsdjeg: This only works in vim 7.4.2010

wsdjeg:
This pr can be improved by 
```viml
filter(map(split(execute('let g:'), "\n"), "matchstr(v:val, '\\S\\+')"), "v:val =~# '^spacevim_'")
```

thanks , @tweekmonster

PR #163, opened 2017-01-25
fix perl completion

wsdjeg:


wsdjeg:
@c9s why can not autocompletion after `->`, ok, I got it, 😭 

https://github.com/c9s/perlomni.vim/blob/master/rplugin/python3/deoplete/sources/perlomni.py#L11

PR #186, opened 2017-01-31
Fix #183

### LINKED ISSUE #183: your vim is too old, getcompletion() need patch7-4-2010
### Expected behavior, english is recommend

I found that the autocomplete for python dose not work, and I have already `$ brew info neovim/neovim/neovim`
### Environment Information
- OS: macOS 10.12.3
- vim version: 
- neovim version: 0.1.7

### The reproduce ways from Vim starting (Required!)

Anytime when I enter the nvim

### Output of the ":message" command, and ":echo SpaceVim#logger#viewLog()"

`:message` cannot display any thing

and `:echo SpaceVim#logger#viewLog()`

```log

SpaceVim Options :

your vim is too old, getcompletion() need patch7-4-2010
[ SpaceVim ] : logger file /Users/limbo/.SpaceVim/.SpaceVim.log does not exists, only log for current process will be shown!

```

wsdjeg:


wsdjeg:
@tweekmonster
Thanks for your code, I just use the func you shown me, but there are two issues.
1. `echo SpaceVim#options#list()[0]`, you will see `:echo SpaceVim#options#listg:spacevim_smartcloseignoreft = ['help']`, but in new version vim, `:echo SpaceVim#options#list` will be clear before echo result.
2. I should support vim version < 7.4.2008, which add `execute()`.

tweekmonster:
@wsdjeg https://github.com/tweekmonster/exception.vim/blob/dedf1c55214cf46b1fc1b4b4d7143666fa90e388/autoload/exception.vim#L98-L108

wsdjeg:
@tweekmonster thanks for the link, it fix the issue 2. but for issue 1, I have no idea, maybe I need `redraw` before output result?

wsdjeg:
@tweekmonster sorry, issue 1 only happen in neovim, I think it is a bug.

tweekmonster:
@wsdjeg np.

You might need to update this to use `'silent let g:'`.  The `:redir` command will still print without `silent`.

wsdjeg:
@tweekmonster how about  change `execute a:cmd` to `silent execute a:cmd`?

tweekmonster:
That's up to you.  I left it out because I think there are some things that aren't captured with `execute()` if you use `silent`.  But, maybe it's fixed now.

wsdjeg:
Thanks , I will update it. 😄

PR #214, opened 2017-02-05
Fix php layer

### LINKED ISSUE #213: I am getting error in php
when I new create file in php framework , I am getting error like following 

## Environment Information
- OS:windows
- vim version:8.0
- neovim version:

wsdjeg:
close #213 

@lvht, I just remove padawan-php/deoplete-padawan in php layer, only phpcd.vim in this layer, but I still get errors , I think it is due to the requirement does not be installed successfully.
so, do you have any idea how to check the requirement?

### DIFF
--- a/autoload/SpaceVim/layers/lang/php.vim
+++ b/autoload/SpaceVim/layers/lang/php.vim
@@ -8,7 +8,8 @@
 " >
 "   PHP 5.3+
 "   PCNTL Extension
-"   Msgpack 0.5.7+(for NeoVim) Extension or JSON(for Vim 7.4+) Extension
+"   Msgpack 0.5.7+(for NeoVim)Extension: https://github.com/msgpack/msgpack-php
+"   JSON(for Vim 7.4+)Extension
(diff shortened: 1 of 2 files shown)


wsdjeg:
What I did?
```
cd ~/src
git clone git@github.com:wsdjeg/msgpack-php.git
cd src/msgpack-php/
phpize
./configure
make
sudo make install
```
out put:
```
Installing shared extensions:     /usr/lib/php/modules/
Installing header files:           /usr/include/php/
```
so I edite php.init
```
extension=msgpack.so
include_path = ".:/php/includes:/usr/include/php/"
```

but  still get no msgpack mode
```
✔ wsdjeg:msgpack-php [ master | …100 ] ➭ php -m | grep msg
PHP Warning:  PHP Startup: msgpack: Unable to initialize module
Module compiled with module API=20151012
PHP    compiled with module API=20160303
These options need to match
 in Unknown on line 0
✘ wsdjeg:msgpack-php [ master | …100 ] ➭ 
```

PR #282, opened 2017-02-21
Tmux fixes

camdencheek:
Made tmux layer work better by default by overriding lazy-loaded plugin mappings. Added a warning that mappings are overrided in the docs. Also added alink to a fix of a known terminfo issue in neovim that misinterprets <C-h> as backspace.

### DIFF
--- a/autoload/SpaceVim/layers/tmux.vim
+++ b/autoload/SpaceVim/layers/tmux.vim
@@ -4,16 +4,19 @@
 " Adds integration between tmux and vim panes. Switch between panes
 " seamlessly.
 " This layer is not added by default. To include it, add
-" `SpaceVim#layers#load('tmux')` to your `~/.SpaceVim.d/init.vim`
+" `SpaceVim#layers#load('tmux')` to your `~/.SpaceVim.d/init.vim`.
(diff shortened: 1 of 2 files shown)


autoload/SpaceVim/layers/tmux.vim
```diff
@@ -24,8 +27,14 @@ endfunction
 
 function! SpaceVim#layers#tmux#config() abort
     let g:tmux_navigator_no_mappings = 1
-    nnoremap <silent> <C-h> :TmuxNavigateLeft<cr>
-    nnoremap <silent> <C-j> :TmuxNavigateDown<cr>
-    nnoremap <silent> <C-k> :TmuxNavigateUp<cr>
-    nnoremap <silent> <C-l> :TmuxNavigateRight<cr>
+    augroup custom_config
```
wsdjeg: It is better to use other group name , because someone maybe use this in custom config file, how about spacevim_layer_tmux

camdencheek: Changed the name in commit 707c8e4. Also changed the name of the script function just for good measure.

Issue #292, opened 2017-02-25
Error Using Shortcuts


NWuensche:
## Expected behavior, english is recommend
Hello everybody, 
I tried to install Spacevim, but everytime I try to e.g. press o for a new line, I get the following error

``
Error detected while processing function <SNR>26_tool:                                       
line    2:                                                                                   
E492: Not an editor command:         IndentLinesToggle   
``

I never had problems while using normal vim. 

## Environment Information
- OS: Arch Linux
- vim version: 8.0
- neovim version: 0.1.7 (But not used)

## The reproduce ways from Vim starting (Required!)
Install Spacevim as told on the home page.
open any file with vim
press 'o'

## Output of the ":message" command, and ":echo SpaceVim#logger#viewLog()"

``
E492: Not an editor command:         IndentLinesToggle
Error detected while processing function <SNR>26_tool:
``

please post log below, if you want me reproduce your issue quickly, post your custom config here will be better.

config file is untouched

wsdjeg:
You need wait for the process of installing plugin.

NWuensche:
Thanks for the answer! But what do you mean with wait? It said that installation was finished.

Edit: I found the missing command in the README. I made a PR to add it also on the website.

PR #293, opened 2017-02-26
Write how to install plugin on Linux on Website.

NWuensche:
As I had the struggle of installing SpaceVim on Linux ( #292 ) , I updated the Website to say that you should install the plugin. I found this line in the README, but not on the website.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -35,6 +35,8 @@ for more info about the install script, please check:
 curl -sLf https://spacevim.org/install.sh | bash -s -- -h
 ```
 
+Before you use SpaceVim, you should install the plugin by executing `:call dein#install()` in (neo-)vim.
+
 ### windows support
 
 - For vim in windows, please just clone this repo as vimfiles in you Home directory.


wsdjeg:
sorry, no need to to run this command after vim startup, SpaceVim will check all the plugins, if there are uninstalled plugin, this command will be called automatically

wsdjeg:
these two lines have been delete from the website in old commit,

NWuensche:
It did not work for me. I had to install it manually.

wsdjeg:
what is the reault of `:echo dein#check_install()`?

NWuensche:
`[dein] Not installed plugins: ['gist-vim']`

wsdjeg:
you can read `:h dein#check_install()`, and if the plugins is not installed it will return no-zero,
and in SpaceVim's scr,
```viml
        if g:spacevim_checkinstall == 1
            silent! let flag = dein#check_install()
            if flag
                augroup SpaceVimCheckInstall
                    au!
                    au VimEnter * call dein#install()
                augroup END
            endif
        endif
```
if there is plugin which is no installed, dein will automatically install these plugins after VimEnter.

NWuensche:
I'm sorry, but I don't understand what you're trying to explain to me regarding the problem.

wsdjeg:
I mean, SpaceVim will auto install all the plugins, and no need to call manually. if not , if should be a bug of dein.

wsdjeg:
merged, will make some change.

PR #359, opened 2017-03-25
Add ui for default manager dein.vim

wsdjeg:
spacevim use dein as default plugin manager, but dein does not have a UI like plug.vim, it is hard for beginner to know how many plugins need to be installed, and what is the status of the installation process.

many people uninstall spacevim just because vim block on first startup. I want to say, it is not block, it is just installing plugins spacevim need. 

now I will implement a UI for spacevim plugin manager.
here is a picture:


FazleArefin:
Exactly. I was almost about to uninstall SpaceVim when I first tried it. 

BTW, why not use `vim-plug` by default instead of `dein`?

wsdjeg:
@fazlearefin dein has many awesome feature SpaceVim need, for example plugin hooks, or merge plugins, I hate long &rtp.

FazleArefin:
@wsdjeg Ok, fair enough.

wsdjeg:
@Shougo I need a func to get a list of uninstall plugin's name, the name should same as the keys in `dein#get()`. I just read src of dein. and find a function.
```viml
function! dein#util#_check_install(plugins) abort
  let plugins = filter(empty(a:plugins) ? values(dein#get()) :
        \ filter(map(dein#util#_convert2list(a:plugins),
        \ 'dein#get(v:val)'), '!empty(v:val)'),
        \     '!isdirectory(v:val.path)')
  if empty(plugins) | return 0 | endif
  call dein#util#_notify('Not installed plugins: ' .
        \ string(map(plugins, 'v:val.name')))
  return 1
endfunction
```
here is my local function:
```viml
function! s:get_uninstalled_plugins() abort
        return filter(values(dein#get()), '!isdirectory(v:val.path)')
endfunction

```

PR #374, opened 2017-03-29
fixes #373 ctrlp configs are all dynamic

### LINKED ISSUE #373: unable to set my own config for plugins
## Expected behavior, english is recommend

I expect to be able to set my own config for plugins... for example:

You have:
`let g:ctrlp_root_markers = 'pom.xml'`
I want:
`let g:ctrlp_root_markers = ['package.json', 'settings.json']`

Ideally - we could have a second "override.vim" file or something, or every one of your configs should be optional, only if not otherwise set.

zeroasterisk:
https://github.com/SpaceVim/SpaceVim/issues/373

I went ahead and just made all of the configs dynamic... vim macros FTW!

If you like that approach - we could do that with all of the other configs too...?


wsdjeg:
yes, you Can do it. thanks.

wsdjeg:
nice work.

PR #375, opened 2017-03-29
re #373 & #374 make all plugin configs dynamic

zeroasterisk:
I did this mostly with a macro, and a bit of hand cleanup - but truthfully I kinda flew through it, and use little of it - please review

### DIFF
--- a/config/plugins/denite.vim
+++ b/config/plugins/denite.vim
@@ -4,7 +4,7 @@ let s:sys = SpaceVim#api#import('system')
 
 
 " denite option
-let s:denite_options = get(g:, 'denite_options', {
+let s:denite_options = {
       \ 'default' : {
       \ 'winheight' : 15,
       \ 'mode' : 'insert',
@@ -14,7 +14,7 @@ let s:denite_options = get(g:, 'denite_options', {
       \ 'direction': 'rightbelow',
       \ 'statusline' : 'false',
       \ 'prompt' : '➭',
-      \ }})
+      \ }}
 
 function! s:profile(opts) abort
   for fname in keys(a:opts)
--- a/config/plugins/denite.vim
+++ b/config/plugins/denite.vim
@@ -4,7 +4,8 @@ let s:sys = SpaceVim#api#import('system')
 
 
 " denite option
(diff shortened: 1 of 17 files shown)


zeroasterisk:
It looks like at least some of this isn't right yet....
ss

wsdjeg:
@zeroasterisk when do you get this errors?

zeroasterisk:
when typing --- I found a problem with some of my `get()` where the keys had `.` in the path, but my getter didn't... reverting those lines....

zeroasterisk:
is this valid?
`let g:deoplete#ignore_sources.rust = get(g:, 'deoplete#ignore_sources.rust', ['omni'])`


if not, i can just revert if they have dot in the key path... if so, I can fix...

wsdjeg:
I will test it now.

wsdjeg:
no, it is not valid.
let g:deoplete#ignore_sources.rust = get(g:deoplete#ignore_sources, 'rust', ['omni'])

wsdjeg:
`rust` is a key in `g:deoplete#ignore_sources`

zeroasterisk:
so this should work:
`let g:deoplete#ignore_sources._ = get(g:, 'deoplete#ignore_sources', '_', ['around'])`

wsdjeg:
no, get() func only support 3 argv
so it should be
`let g:deoplete#ignore_sources._ = get(g:deoplete#ignore_sources,  '_',  ['around'])`

zeroasterisk:
^ that seems to have fixed it - from what I can see

config/plugins/denite.vim
```diff
@@ -13,7 +14,7 @@ let s:denite_options = {'default' : {
       \ 'direction': 'rightbelow',
       \ 'statusline' : 'false',
       \ 'prompt' : '➭',
-      \ }}
+      \ }})
```
wsdjeg: no need change s:value, this will not be changed by users.

zeroasterisk: reverted

wsdjeg:
LGTM, 👍

PR #383, opened 2017-03-31
Set guifont

wsdjeg:


wsdjeg:
will look into this PR. SpaceVim should support other fonts

PR #421, opened 2017-04-09
Fix locale setting on SunOS

jen20:
SunOS kernels use the `en_US.UTF-8` style of locale definition. This modification checks the `uname` against `SunOS\n` and sets the language appropriately.

After applying this patch, SpaceVim works perfectly on SmartOS.


wsdjeg:
Thanks.

PR #423, opened 2017-04-10
add vim-scala plugins in scala lang layer for synctax highting and more

icearith:


autoload/SpaceVim/layers/lang/scala.vim
```diff
@@ -0,0 +1,14 @@
+""
+" @section lang#scala, layer-lang-scala
+" @parentsection layers
+" @subsection mappings
+" >
+"   mode            key             function
```
wsdjeg: If there is no mapping, you can remove these three lines.

PR #436, opened 2017-04-14
Use on_event, instead of deprecated on_i

jetm:
Based on dein plugin documentation, 'on_i' or 'dein-options-on_i' is deprecated. It should be used 'on_event'.

'on_i' will call dein#source() on InsertEnter autocmd event, and the equivalent for 'on_event' is "'on_event' : 'InsertEnter'".

### DIFF
--- a/autoload/SpaceVim/layers/autocomplete.vim
+++ b/autoload/SpaceVim/layers/autocomplete.vim
@@ -22,18 +22,18 @@
 
 function! SpaceVim#layers#autocomplete#plugins() abort
   let plugins = [
-        \ ['honza/vim-snippets',          { 'on_i' : 1, 'loadconf_before' : 1}],
-        \ ['Shougo/neco-syntax',          { 'on_i' : 1}],
-        \ ['ujihisa/neco-look',           { 'on_i' : 1}],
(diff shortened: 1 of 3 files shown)


autoload/SpaceVim/layers/autocomplete.vim
```diff
@@ -22,18 +22,18 @@
 
 function! SpaceVim#layers#autocomplete#plugins() abort
   let plugins = [
-        \ ['honza/vim-snippets',          { 'on_i' : 1, 'loadconf_before' : 1}],
-        \ ['Shougo/neco-syntax',          { 'on_i' : 1}],
-        \ ['ujihisa/neco-look',           { 'on_i' : 1}],
-        \ ['Shougo/context_filetype.vim', { 'on_i' : 1}],
-        \ ['Shougo/neoinclude.vim',       { 'on_i' : 1}],
+        \ ['honza/vim-snippets',          { 'on_event' : 1, 'loadconf_before' : 1}],
```
wsdjeg: if you use on event as key, the value should be event name instead of 1.

jetm: Thanks, @wsdjeg. Let me try to figure out the right dein events. No much experience with dein and VimL.

PR #459, opened 2017-04-22
Add airline support for dein ui

wsdjeg:
https://github.com/vim-airline/vim-airline/issues/1464


wsdjeg:
when reopen plugin manager ui buffer , I got warning
`The airline statusline funcref function('MyPlugin') has already been added.`

the ftplugin should not be loaded again!

PR #465, opened 2017-04-23
Add APIs

L-stt:
Most of these funcs are from vital
- [x] vim#message
- [x] vim#buffer
- [ ] vim#buffer_manager
- [ ] vim#search
- [ ] time

### DIFF
--- a/test/api/vim/buffer.vader
+++ b/test/api/vim/buffer.vader
@@ -1,9 +1,6 @@
 Execute ( SpaceVim api: vim#buffer ):
   new
   let buffer = SpaceVim#api#import('vim#buffer')
-  normal! g:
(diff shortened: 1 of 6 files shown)


wsdjeg:
Nice work, thanks!

test/api/vim/buffer.vader
```diff
@@ -3,5 +3,7 @@ Execute ( SpaceVim api: vim#buffer ):
   let buffer = SpaceVim#api#import('vim#buffer')
   normal! g:
```
wsdjeg: here should be `call feedkeys('q:', 'n')`.  and issue is opened here : https://github.com/junegunn/vader.vim/issues/129

PR #476, opened 2017-04-28
Add more lovely banners

Gabirel:
Why not just make SpaceVim `call SpaceVim#layers#load('core#banner')` actively?


Gabirel:
Done.
Two suggestions:
- [x] `call SpaceVim#layers#load('core#banner')` actively. Many users won't take too long to read documents, they may never find this lovely feature.
- [ ] According to the width of different logos and the width of window, select appropriate logo to load. So user may not encounter a mess when start using vim

Gabirel:
@wsdjeg I just finished the first one.

However I have no confidence in the second one

You could merge it right now or just implement this feature by yourself. Because I don't know how to do it.

Or, just ignore this feature

wsdjeg:
I will implement the second feature.

PR #482, opened 2017-05-01
prevent overriding if g:startify_custom_header is set

techgaun:
This seems to be prevent the ability to use the `g:startify_custom_header` that I specified.. What do you think of this change so users can still specify their own `g:startify_custom_header`?


wsdjeg:
This PR will disable the refresh of welcome page logo. now every time you run `:Startify` for press `Space + a + s`, the logo in welcome page will be changed randomly.

techgaun:
sorry yeah @wsdjeg I didn't think about that at all

PR #492, opened 2017-05-05
only call update logo if g:startify_custom_header does not exist

techgaun:
hey @wsdjeg based on your feedback on #482 (sorry, closed that one accidentally), what do you think of calling `s:update_logo()` only if `g:startify_custom_header` does not exist. This still allows using `Space+a+s`.


wsdjeg:
in autocmd.vim  there is a autocmd call this func. that line should also be surrounded by the checker.

Issue #516, opened 2017-05-11
SPC-f-f/SPC-p-f bindings always say "NO ENTRIES"


reem:
## Expected behavior, english is recommend

The files in the current directory/project respectively are listed.

## Environment Information
- OS: macOS 10.12
- vim version: 8.0
- neovim version: 0.2.0

## The reproduce ways from Vim starting (Required!)

Just open any file in this repo (e.g. README.md) and try SPC-f-f or SPC-p-f.

## Output of the ':SPDebugInfo!'

Same as in https://github.com/SpaceVim/SpaceVim/issues/515

## Screenshots

Here's an asciinema recording of the issue using this repository as an example: https://asciinema.org/a/3hevp4ew9o7md1twl8dmjsuij

wsdjeg:
I can not reproduce it,.

https://asciinema.org/a/7a1qj8d3dizxjus95zfyq61x0

wsdjeg:
ok. I got it, your vim has no python support,  you should use recompile your vim.

wsdjeg:
will fix it

wsdjeg:
It in config of ctrlp
```vim
if !exists('g:ctrlp_match_func') && (has('python') || has('python3'))
  let g:ctrlp_match_func = { 'match': 'pymatcher#PyMatch'  }
endif
```
so, I think SpaceVim has no bugs, if your vim do not support py, the pymatch will not be loaded, so It should not be `NO ENTRIES`.

wsdjeg:
I can not reproduce in old vim:
```
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled May 15 2017 19:38:26)
Included patches: 1
Compiled by wsdjeg
Huge version with GTK2 GUI.  Features included (+) or not (-):
+arabic          +file_in_path    +mouse_sgr       +tag_binary
+autocmd         +find_in_path    -mouse_sysmouse  +tag_old_static
+balloon_eval    +float           +mouse_urxvt     -tag_any_white
+browse          +folding         +mouse_xterm     -tcl
++builtin_terms  -footer          +multi_byte      +terminfo
+byte_offset     +fork()          +multi_lang      +termresponse
+cindent         +gettext         -mzscheme        +textobjects
+clientserver    -hangul_input    +netbeans_intg   +title
+clipboard       +iconv           +path_extra      +toolbar
+cmdline_compl   +insert_expand   -perl            +user_commands
[...]

PR #538, opened 2017-05-15
Fix exclude syntax (-g) for rg util

jetm:
A fix for #516 issue.

In rg values given to -g must be quoted or your shell will expand them
and result in unexpected behavior. For example: rg -g '\<glob\>'.


wsdjeg:
@jetm This PR break denite config, I need to fix it.

PR #549, opened 2017-05-21
Add lang#lisp layer

wsdjeg:
- [x] add layer doc

- [x] add language spcified mapping `[SPC] l  *`


wsdjeg:
~~I open a PR to vlime repo, and will make some udpate for this layer.~~

ok, we can use server manager.

PR #579, opened 2017-05-30
Improve gitcommit completion

wsdjeg:


wsdjeg:
@Shogo, I want to implement fuzzy match github issue number and pr number, here is the video, but it can only match number, can not match title.
https://asciinema.org/a/3r900qipvhrylt4jkbrks2n6s

PR #670, opened 2017-06-26
Add 'on_ft':'rust' for rust.vim

wy15:
rust.vim not load


wsdjeg:
@xiaobagou vim has not ftdetect for rust file. if you make this plugin loaded on_ft:rust, then when you open a *.rs file the filetype will not be rust.

BTW, we can merge this PR, just to make vim startup more fast, but we need to add
```
au BufRead,BufNewFile *.rs set filetype=rust
```
to https://github.com/SpaceVim/SpaceVim/blob/dev/filetype.vim

PR #688, opened 2017-06-29
Add CompleteParameter when use deoplete as the completion engine.

tenfyzhong:


wsdjeg:
@tenfyzhong https://asciinema.org/a/sC6fTdwvutkjgbwMKHPa86vZS This is default behavier of neosnippet, I not sure if this plugin will conflict with it.

Shougo:
If `g:neosnippet#enable_completed_snippet` is 1, it will conflict.

wsdjeg:
I can not merge this PR now, because in spacevim `tab` is mapped to smart_tab function, and in this func we need to check many things, for example:
1.  argv_jump_able?
2.  snippet_expand_able?
3.  snippet_jump_able?
4, we also need to check the character under the cursor, before the cursor and after the cursor.

tenfyzhong:
>argv_jump_able?
snippet_expand_able?
snippet_jump_able?
we also need to check the character under the cursor, before the cursor and after the cursor.

bc71ec9 test success.

PR #717, opened 2017-07-09
Close job/timer when close flygrep buffer

### LINKED ISSUE #716: Fly grep timer and job should be closed when press Esc.
I just read the code of fly grep plugin, and I think the job/timer should be closed when user press Esc. if you just quit buffer, the previous buffer will be changed by the job's stdout callback func.

Zym-xx:
close  #716


wsdjeg:
Thanks!

PR #730, opened 2017-07-17
Fix typo: s/recommenced/recommended/

kenjitoyama:
I believe "recommenced" is not right here and the intended word was "recommended."


wsdjeg:
yeah. thanks.

PR #765, opened 2017-08-01
Fix issue with n vs N when ignorecase is set

mdavezac:
If ignorecase is set, then string comparisons in vimscript ignore cases.
Hence `if a:key == 'n'` is true when `key` is 'N'.


wsdjeg:
oh , it is not issue which latter you use. What need to do is just change `==` to `==#`

wsdjeg:
thanks

mdavezac:
Thanks, I didn't know about vimscripts' ==#

PR #811, opened 2017-08-29
Add buffer # keymap

zhujinxuan:
Add keymap `SPC b #` to command `buffer #` to open the previously opened buffer.  I think it is quite a useful command especially when you open multi-buffers but temporally focus on two buffers.

### DIFF
--- a/autoload/SpaceVim/mapping/space.vim
+++ b/autoload/SpaceVim/mapping/space.vim
@@ -86,6 +86,8 @@ function! SpaceVim#mapping#space#init() abort
         \ ]
         \ ]
         \ , 1)
+  call SpaceVim#mapping#space#def('nnoremap', ['b', '#'], 'buffer #', 'Previously active buffer (:b #)', 1)
+
   "
   " Comments sections
   "


Motardo:
Hi @zhujinxuan, Is this different from <kbd>Space</kbd><kbd>Tab</kbd> in the code below?
[https://github.com/SpaceVim/SpaceVim/blob/11c7d5cbdaa7a2d5defcde6b6084930a8d781022/autoload/SpaceVim/layers/default.vim#L118](https://github.com/SpaceVim/SpaceVim/blob/11c7d5cbdaa7a2d5defcde6b6084930a8d781022/autoload/SpaceVim/layers/default.vim#L118)

zhujinxuan:
@Motardo not different.  Sorry I did not notice that <tab>.

wsdjeg:
@zhujinxuan I think `SPC b #` is hard too press, we need to press `SPC b Shift + 3`, It is hard than `:b#`, also hard than `SPC tab`. 

but `SPC b #` is  easy to understand, we can keep this PR open, and if there are some other people also like this mapping, we can merge it.

PR #814, opened 2017-08-30
Add lang#julia layer

zhujinxuan:
Add [Julia Language](https://julialang.org/) support, with syntax and lint.  The completion only works in neovim due to deoplete.


wsdjeg:
Nice,thanks.

PR #823, opened 2017-09-05
Add layer for Pony programming language

oldstager:


wsdjeg:
how about https://github.com/jmcomets/vim-pony

any plan to include it?

crisidev:
@wsdjeg, the plugin you posted is not for po ylang, but for Django and python.

bpdp:
yes @crisidev is right. This PR is for https://www.ponylang.org

wsdjeg:
Thanks, I just test this layer, as @oldstager include this plugin with `on_ft` option, so only when filetype is change to pony, the plugin will be loaded, but before this plugin is loaded, the ftdetect is not work. I think we should make the ftdetect script not lazy loed.

wsdjeg:
@Shougo how does dein deal with `ftdetect` script, when we set `on_ft` option?

PR #851, opened 2017-09-17
`eslint --fix` for javascript

zhujinxuan:
`eslint --fix` can auto fix common linting problems and style the javascript code.  I think it would be good if we turn on it by default rather than forcing users calling them by `!eslint --fix %` by hand.  

Here I contribute a config in my `init.vim` to turn on the `eslint --fix`; and for anyone who want to disable `eslint --fix`, they can add `let g:spacevim_javascript_autofix = 0` before loading the javascript layer.

Thanks a lot for `SpaceVim`~

### DIFF
--- a/autoload/SpaceVim/layers/lang/javascript.vim
+++ b/autoload/SpaceVim/layers/lang/javascript.vim
@@ -17,6 +17,17 @@ endfunction
 
 function! SpaceVim#layers#lang#javascript#config() abort
   call SpaceVim#mapping#gd#add('javascript', function('s:gotodef'))
+  if !exists(g:spacevim_javascript_autofix)
+    let g:spacevim_javascript_autofix = 1
+  endif
+  if (g:spacevim_javascript_autofix) 
+    " Only use eslint
+    let g:neomake_javascript_enabled_makers = ['eslint']
+    " Use the fix option of eslint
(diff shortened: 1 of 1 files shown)


wsdjeg:
@zhujinxuan I think this is same feature as I used in markdow layer, in markdown layer, we use [remark](https://github.com/wooorm/remark) as the default text format tools, it can format the content of current file, also it can fix style.

so I think maybe add a new feature: ` fix content when formatting current file` will better.

BTW I am not sure if `eslint --fix` will make any unwanted changes.

zhujinxuan:
`eslint --fix` sometimes makes unwanted changes, unfortunately.  The reason is that some js symbols are reused for different meanings, like {} for code block and for object.  However, it is useful when write js scripts. 

I think we can turn off with `g:spacevim_javascript_autofix = 0` by default and add document that: you can turn on it by `g:spacevim_javascript_autofix = 1` before loading the layer.

wsdjeg:
How about use lang specifed option? then we can load this layer via:

```
call SpaceVim#layers#load('lang#javascript', {
       \ 'autofix' : 1,
       \ })
```

zhujinxuan:
That is great. But i do not know how to detect 'autofix' in the vimscript.

wsdjeg:
ok, I will implement the feature base on this PR.

zhujinxuan:
Thanks ~

PR #859, opened 2017-09-20
[WIP] - Implement Sass files

lsrdg:
This pull request makes use of jekyll's native support to SASS. It is just a **minimal** sample, more `$variables` could be implemented as needed.

Just as a side note: as far as I could tell (after my quick contact with Midnight's code), there are specific styles to `h1`, `h2` an `h3` headers, but not to `h4`, `h5` and `h6`, what could explain the problem to customize  `h4`, `h5` and `h6`.


wsdjeg:
@lsrdg maybe I should not use all of `h1` ... `h6` in one page, it is hard to read. I think the github's markdown parser is better, here is a example

# TEST H1

## TEST H2

### TEST H3

#### TEST H4

##### TEST H5

###### TEST H6

lsrdg:
> not use all of h1 ... h6 in one page, it is hard to read.

Totally agree with that.

> I think the github's markdown parser is better

Sorry, I am not sure about what is that supposed to mean. And the output is just the six headers.
Just to be clear, when you have a header, AFAIK, it will be rendered by jekyll, despite of it being pure html or markdown.

wsdjeg:
in the website, we use `H1` as a logo, on the top of the website, if we change the font size of H1, the logo looks different from old one.

wsdjeg:
@lsrdg ohh, I am sorry for late reply, I want to merge this branch, but it has been removed from network. and if is it possiable to custom all the headers, (from H1 to H6)?

lsrdg:
Sorry, it was me who missed your [comment](https://github.com/SpaceVim/SpaceVim/pull/859#issuecomment-332208487).

> I want to merge this branch, but it has been removed from network. 

I don't get it. Should I make a new pull request with that?

> if is it possible to custom all the headers, (from H1 to H6)?

You do want to customize the H4, H5 and H6? If so, that should probably be done with caution, to avoid incompatibility, specially in case that the midnight theme gets some updates.

wsdjeg:
Thanks. we I also think we should avoid incompatibility.

and in the longest page (documentation), we use H1 - H5 as I know. at least we should be able to custom these headers.

A new PR is welcome.

lsrdg:
Ok. Just to be clear: the new pull request you need is the same as #859 ?
Because I would rather make another pull request to support the headers. What do you think?

wsdjeg:
both is OK to me.

wsdjeg:
OK I will try to merge this.  and you  just need to featch the remote branch.

lsrdg:
So, just to be clear: #859 is now #1083.
About the headers support, I have one last question before closing this issue: are you pulling in changes from the midnight theme? It makes more sense (at least to me) to add support to the headers directly to the theme repository, so everyone would benefit from it.

wsdjeg:
what do you mean 'are you pulling in changes from the midnight theme'? I think I am using the latest version of midnight, because it is installed by Gemfiles. but I just change the default.html, which is in docs/_layout/

lsrdg:
Ok. I'll see if the midnight repo wants it. If it does, I will send the PR there and SP will get it (since you are using the gem), orelse I'll just send it here. Either way, I'll let you know.

wsdjeg:
Thanks😊

PR #872, opened 2017-10-04
[WIP] Improve tests

blueyed:
Includes #881.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,5 +58,8 @@ script:
     elif [ "$LINT" = "vint-errors" ]; then
       vint --error .
     elif [ "$LINT" = "vader" ]; then
-      make test
+      make test_coverage
+      build/covimerage/bin/covimerage xml
+      pip install codecov
+      codecov -X search gcov pycov -f coverage.xml
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ script:
     elif [ "$LINT" = "vader" ]; then
       pip install --no-cache-dir https://github.com/Vimjas/covimerage/archive/develop.zip
       make test_coverage
-      covimerage xml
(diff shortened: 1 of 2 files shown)


blueyed:
You have some coverage now at https://codecov.io/gh/SpaceVim/SpaceVim/pull/872/tree.. :fireworks:

blueyed:
It currently only report files actually being executed, so your whole coverage is even worse than the 37% - but that is fair considering that you only have less than a handful of tests.

As for covimerage, it should maybe include all the .vim files based on its own glob, so that it also includes not executed files.

wsdjeg:
great work! thanks.

wsdjeg:
@blueyed Thanks for your works, I just see there are two checks failed, is it due to SpaceVim's tests are too little?

blueyed:
@wsdjeg 
The checks are expected, since there is nothing to report against.
Later you will see there if coverage increased and if the diff coverage is good etc.
I have something in mind to improve the coverage reporting, by including all source files etc.. I will update this PR later.

For the tests itself, you might want to consider using Vim in Docker (via https://github.com/tweekmonster/vim-testbed, also used for Neomake).
And then it would make sense to use CircleCI instead of Travis, since they have good Docker support (where you can sepcify the image for each job), and it is really fast and more pleasant to use.

blueyed:
This looks good from my side.
See https://codecov.io/gh/SpaceVim/SpaceVim/tree/e1c4cac000a462828b8345f316803feb5ac842a8 to browse the files.
Maybe you want to omit more, e.g. `SpaceVim.d`, not sure.

Not sure what your plan is, but in general codecov/this can help to bring the coverage up, e.g. if you at least cover the new code etc.

wsdjeg:
@blueyed Thanks, I think it is ok now, the next step I need to do should be add some test for spacevim.

PR #879, opened 2017-10-05
Travis: set -e with script blocks

blueyed:
This makes it fail actually if there are any errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ matrix:
 
 install:
   - |
+    set -e
     if [ "${LINT#vimlint}" != "$LINT" ]; then
       git clone --depth=1 https://github.com/syngan/vim-vimlint /tmp/vimlint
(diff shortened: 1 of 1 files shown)


blueyed:
You can see here that vimlint is actually failing, but it was not caught.

Side note: for Travis it should exit with 1 to cause a ailure, anything else (like this) is an error.
Errors are meant for when something in the setupp failed, but this should be a failure.

wsdjeg:
vimlint-errors is failed. the log is:

```
1Vim(let):E132: Function call depth is higher than 'maxfuncdepth'
```

I will have a look tonight.

wsdjeg:
@blueyed this is due to there are too many vim files, we need to split the command.
issue is opened here : https://github.com/syngan/vim-vimlint/issues/107

blueyed:
You could only run it for changed files - reminded me of https://github.com/neomake/neomake/pull/1651 (see the old `travis_lint` in there).

PR #880, opened 2017-10-06
Travis fix

wsdjeg:
merge #879


blueyed:
The vimlint/vimlint-errors job now take almost 10mins each.

Maybe they should be just split in half - instead of per file?!

wsdjeg:
@blueyed before this PR is merged, I just try to use test command like:

https://github.com/SpaceVim/SpaceVim/pull/880/commits/793f7f16e5b7b08c882d52694aa6e691326d2bb1

but this still get same errors, I think maybe vimlint should be improved.

PR #891, opened 2017-10-10
Test changed vim scripts only

wsdjeg:


wsdjeg:
@blueyed In this PR, I just change the test command, and only parser the change vim files. That will make the test more quick.

PR #905, opened 2017-10-19
Fix a typo in doc/SpaceVim.txt

felixonmars:


wsdjeg:
Thanks for your PR. But SpaceVim use vimdoc to generate doc files. so you need to edit the Vim file instead of txt file.

I just check the diff file. you need to edit autoload/SpaceVim.vim

wsdjeg:
Thanks!

PR #907, opened 2017-10-19
Display mode in statusline like in some airline theme

adrnbttr:
This feature is to offer the possibility to displaying the current mode in winnr section like in some airline theme.


wsdjeg:
nice work。

adrnbttr:
The CLA issue is fixed.

PR #927, opened 2017-10-23
Minor fix for Vim help language setting example

sei40kr:
* Minor fix for Vim help language setting example (`chinese` -> `cn`)


wsdjeg:
nice catch!!

PR #939, opened 2017-10-26
Added a new plugin thinca/vim-quickrun

sei40kr:
Added a new plugin thinca/vim-quickrun.
May conflicts with `vimwiki` PR.


wsdjeg:
@sei40kr sorry, I do not want to use quickrun, SpaceVim has build-in runner, and I will use job/async feature instead.

Issue #925, opened 2017-10-23
Ctrl+J in normal mode gets overridden by vim-tmux-navigator


sei40kr:
## Expected behavior, english is recommended

Press `Ctrl+J` in normal mode -> Search the word under caret in current buffer
But with `tmux` layer, Vim tries to go to a buffer/pane below.

## Environment Information

- OS: macOS & Ubuntu 17.04
- vim version: 
- neovim version: 0.2.0

## The reproduce ways from Vim starting (Required!)

Install both layer `edit` and `tmux` .

## Output of the ':SPDebugInfo!'

please post log below, if you want me to reproduce your issue quickly, including your custom config here will be better.

## Screenshots

If you have any screenshots for this issue, please upload here. BTW you can use https://asciinema.org/ for recording video in terminal.

wsdjeg:
@sei40kr I want to use `ctrl_n` and `ctrl_m` for multiple-cursors, and `ctrl_hjkl` for tmux layer.

wsdjeg:
fixed.

PR #945, opened 2017-10-27
resolve conflicts between tmux layer and edit layer

ssfdust:
https://github.com/SpaceVim/SpaceVim/issues/925
To resolve the conflicts, I subsitute the old key binding 'c-j' and 'c-k' with 'c-n' and 'c-u' which relate to multiple-cursors plugin.On the ohter hand, I replace the key binding 'c-l' with 'c-\_'(means C-/) relating to consult the youdao online dictionary.
The default key binding metioned above for vim or neovim lists below:
C-n move down one line
C-u up half screen
C-_  unbound

### DIFF
--- a/autoload/SpaceVim/layers/edit.vim
+++ b/autoload/SpaceVim/layers/edit.vim
@@ -29,8 +29,8 @@ function! SpaceVim#layers#edit#plugins() abort
 endfunction
 
 function! SpaceVim#layers#edit#config() abort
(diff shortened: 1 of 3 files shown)


wsdjeg:
I think use `ctrl+u` maybe is not good idea, maybe we can use `ctrl+n` and `ctrl+m` for multiple-cursors, as `m` and `n` are very close。 and both of them is just mean `move down one line` which can easy got via `j` `<cr>` or `<down>`.

ssfdust:
Got it.

PR #970, opened 2017-11-07
Fix capitalization on website

Phlosioneer:


Phlosioneer:
The codecov bot's stats make absolutely no sense. I have no idea what it's supposed to do, but its numbers for dev vs 970 should be tiny. Instead it sees two fewer files, 73 lines less, if I'm reading it right?

It doesn't matter to me, just curious.

wsdjeg:
This shoul be issue with codecov itself.

PR #1000, opened 2017-11-14
Added autocmd to update status-line

sei40kr:
# Added autocmd to update status-line

Added autocmd to update status-line on ALE process completion.

This is still not perfect fix.
It needs to detect to the target buffer is currently active or not.


sei40kr:
In short, "Fixed an issue that statusline won't be updated after ALE process"

wsdjeg:
Thanks.

PR #1009, opened 2017-11-17
Add mappings for git blame and git log

crisidev:


wsdjeg:
for git log,

I think the existing mappings `SPC g v` and `SPC g V` is better, it use https://github.com/cohama/agit.vim

crisidev:
@wsdjeg I did not noticed it.. Sorry about that. Would you like to keep the git blame shortcut?

wsdjeg:
I am ok. but you can not use `Gina blame`, maybe it should be `Gina blame :`

lambdalisue:
> I am ok. but you can not use Gina blame, maybe it should be Gina blame :

Recently it changed. So `:Gina blame` is OK for now.

https://github.com/lambdalisue/gina.vim/commit/7358ccabe6c21b934aa597506747313576a5d045

wsdjeg:
Thanks for all of your work.

PR #1027, opened 2017-11-23
uninstall section added

harshitanand:
Uninstall section added to readme


wsdjeg:
this has been added in wiki and website. the install script support many options. user should run with -h instead. BTW, after install spacevim, I recommend to read the doc.

PR #1040, opened 2017-11-29
Enable syntax error/warning counts in status line

mdavezac:
For some reason, unless I'm mistaken, showing the number of errors in the status line is implemented but not enabled.

More specifically, this PR adds 'syntax checking' to the registered status line sections.
The implementation for syntax checking is the one that existed already.


wsdjeg:
Thanks, nice catch!

PR #1054, opened 2017-12-01
get_complete: compatiable with macOS.

tracyone:
Install bash-completion in macOS:

brew "bash-completion"


wsdjeg:
@tracyone can you please sign the cla, thanks.

wsdjeg:
@brbsix @tracyone 

hi, Thanks for you bash completion script, and I use it for vim's omni func, but as you can see it in this pr, some fixes need to be done, but this file is under lgpl-3.0 License, we can only use it without modified.

tracyone:
@wsdjeg You are right.

PR #1064, opened 2017-12-03
zsh:support autocomplete

tracyone:
1. autotrigger in YCM and deoplete.
2. press ^x^o in other case


wsdjeg:
Nice plugin!

tracyone:
@wsdjeg for ycm, https://github.com/Valodim/vim-zsh-completion/pull/2  this pull request solve the issue of autotrigger.

Issue #1081, opened 2017-12-06
Ctrlp does not work.


fabiogibson:
## Expected behavior

Expect to be able to use CtrlP with ag searcher.

## Environment Information
- OS: Arch
- vim version: 8.0
- shell: zsh

## The reproduce ways from Vim starting (Required!)

CtrlP with ag searcher does not execute.

## Output of the ':echo g:ctrlp_user_command'

> ag --hidden -i  -g "" --ignore */tmp/* --ignore *.so --ignore *.swp --ignore *.zip --ignore *.class --ignore tags --ignore *.jpg --ignore *.ttf --ignore *.TTF --ignore *.png --ignore */target/* --ignore .git --ign\nore .svn --ignore .hg --ignore .DS_Store --ignore *.svg %s

Executing this command replacing %s for any valid pattern throws the following:

>  zsh: no matches found: */tmp/*

My system requires the patterns to be enclosed in quotes:

> Ex: --ignore '*.sp'.

I've fixed this replacing the line 280 at /autoload/zvim/util.vim 

from:
> call add(ignore, ig)

to:

> call add(ignore, "'" . ig . "'")

Please let me know if a PR is welcome.

wsdjeg:
@fabiogibson Nice catch, there is [old issue](https://github.com/SpaceVim/SpaceVim/issues/516) talk about ctrlp, but that only fix rg support, and your patch I think will fix ag support, PR welcome, and I will test it in `bash`.

fabiogibson:
Fixed #1082

PR #1082, opened 2017-12-06
Fix for ctrlp with ag searcher. #1081

fabiogibson:
This patch was tested on zsh shell.


wsdjeg:
Thanks, tested in bash, it works well, merged.

PR #1094, opened 2017-12-09
Add language server configurations for Haskell

sei40kr:
# Changes

* Add language server commands for Haskell
* Update key mappings for Haskell

This PR includes another PR (the one which implements JavaScript LSP support)


wsdjeg:
Thanks, I found the test failed every time. I think the test has some bugs, I will look into it later, But I am too busy this weekend, sorry.

PR #1105, opened 2017-12-10
Improve implementations of tmux integration

sei40kr:
## Changes

* Fix an issue that the Vim screen disappears on splitting tmux
* Toggle visibility of cursor line on entering/leaving tmux pane
* Load tmux plugins lazily for faster start-up


wsdjeg:
Nice,merged.

PR #1111, opened 2017-12-11
Improve Emmet configuration

sei40kr:
## Changes

* Use JSX Emmet presets for Javascript

## Before

In JavaScript file,

`div.something` -> `<div class="something"></div>`

## After

`div.something` -> `<div className="something"></div>`


wsdjeg:
Thanks, :smile:

PR #1116, opened 2017-12-13
add doc of searching files to managing project

### LINKED ISSUE #1115: Improve the documentation of searching
Technically, searching in Spacevim can be roughly subdivided into two categories: **searching files** and **search text**.  

Currently, the documentation contains a loads of key mapping that related to searching. I feel like it's ok to have different combinations of key mapping to trigger different searching actions, but keep the two **category** separated is necessary to keep the documentation readable and understandable. 

Apart from that, key mappings like:

1.  `[SPC] p f`: search files in current project is not included in the doc, which is super useful then.
2. `[SPC] s g G`: grep text  on the fly in current project has two different descriptions in the doc, which should be improved to keep consistency. 

I would suggest that we should split the documentation of searching text and searching file. If you guys think this is necessary, I will send a new PR.

icearith:
close #1115

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -93,6 +93,7 @@ description: "General documentation about how to using SpaceVim, including the q
     - [Multi-Encodings](#multi-encodings)
   - [Errors handling](#errors-handling)
   - [Managing projects](#managing-projects)
+    - [Searching files in project](#searching-files-in-project)
 - [EditorConfig](#editorconfig)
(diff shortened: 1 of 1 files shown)


wsdjeg:
Thanks!

icearith:
@wsdjeg  I am a beginner of Vim developer, so I will start with simple things and gradually move on, thanks for your help. Long live Spacevim 👍

PR #1149, opened 2017-12-20
Fix lua repl

wsdjeg:


icearith:
@wsdjeg, I just realised that I created an alias from `luap` to `luaprompt` several months ago.

wsdjeg:
Hmm， I also get issue about the lua layer.

https://github.com/SpaceVim/SpaceVim/issues/1150

PR #1200, opened 2017-12-29
[SPC b t] Open Vimfiler by buffer file dir

zhujinxuan:
By default, `:Vimfiler` opens file tree at project dir, but sometimes we may prefer to look around (for example by "../" or "child/" at current buffer file dir).  Therefore I add a key binding under `SPC b` for that.


wsdjeg:
nice. by why need -no-toggle? I think it should keep same behavior as SPC f t.

zhujinxuan:
No-toggle provides this behavior:
If vimfiler tree is opened, jump the file tree to the buffer for.

wsdjeg:
I do not know this feature. I will test it. Thanks for your pr.

PR #1257, opened 2018-01-08
Option to use local eslint 

zhujinxuan:
A design problem of eslint is that globally installed eslint cannot use the eslint-rules defined in the locally installed packages.  Therefore, when working with multiple js projects, local eslint is often sued for different rules in different packages.  

Here I add an option as `use_local_eslint` for use eslint from locally installed `node_modules/.bin/eslint`

Cheers,
Jinxuan Zhu


wsdjeg:
Thanks. merged!

PR #1306, opened 2018-01-21
Floating windows for flygrep

SpaceVimBot:
reopen #1102


wsdjeg:
```
git fetch upstream  refs/pull/1306/head:foo
git checkout foo
git push -u origin foo
```

PR #1359, opened 2018-01-31
Add prompt for quit buffer.

wsdjeg:
fix_quit

### DIFF
--- a/autoload/SpaceVim/mapping.vim
+++ b/autoload/SpaceVim/mapping.vim
@@ -143,7 +143,19 @@ endfunction
 function! SpaceVim#mapping#close_current_buffer() abort
   let buffers = get(g:, '_spacevim_list_buffers', [])
   let bn = bufnr('%')
-  let index = index(buffers, bn) 
+  if getbufvar(bn, '&modified', 0)
+    redraw!
+    echohl WarningMsg
+    echon 'current buffer contains unsaved modification, save file Y/N?'
+    echohl None
+    let rs = nr2char(getchar())
+    if rs ==? 'y'
+      write
+    else
+      return
+    endif
+  endif
+  let index = index(buffers, bn)
   if index != -1
     if index == 0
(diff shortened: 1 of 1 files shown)


autoload/SpaceVim/mapping.vim
```diff
@@ -143,7 +143,23 @@ endfunction
 function! SpaceVim#mapping#close_current_buffer() abort
   let buffers = get(g:, '_spacevim_list_buffers', [])
   let bn = bufnr('%')
-  let index = index(buffers, bn) 
+  if getbufvar(bn, '&modified', 0)
+    redraw!
+    echohl WarningMsg
+    echon 'current buffer contains unsaved modification, save file Y/N?'
```
icearith: what will happen if a buffer is not opened from a file?

wsdjeg: It only check if the buffer is modified，even if it is not opened from  a file.

icearith: yap, I am wondering if the file is not opened from a file, how to save it? does the user need to manually input a file name?

tracyone:
You just need to 

```

set confirm

```

wsdjeg:
this is enabled by default，but when press key binding, the error is ugly.

you can see the pic in issue

tracyone:
Weird.

tracyone:
With following minimal config,

```vim
 set nocompatible
set confirm
colorscheme desert "default setting 
filetype plugin indent on
syntax on

```

screen shot 2018-02-04 at 11 21 32

wsdjeg:
But I am not run `:q`, I am calling a func, in this func will will use `:bd`, this will print the function name when get errors. and I do not want to print the function name.

BTW, I hate two line message, it will push my screen up.

icearith (approved):
I feel like this PR can be merged into master.

PR #1389, opened 2018-02-07
Add icon to welcome page(startify)

wsdjeg:
ref: https://github.com/mhinz/vim-startify/pull/299


icearith (commented):
should we also add ICON to vimfiler

wsdjeg:
that has been done in SpaceVim

Issue #1436, opened 2018-02-23
Discussion about better default


wsdjeg:
I want to make SpaceVim out of box. which means The default config should be only needed by all users.

sei40kr:
https://github.com/sei40kr/dotfiles/blob/master/_SpaceVim.d/init.vim
This is my SpaceVim config.

* Search Ruby and Python in asdf, pyenv to make sure that SpaceVim uses user-installed versions.
* Change the leader key to `;` because it's comfortable on JP layout keyboard.
* Search libclang installed via Homebrew/Linuxbrew to avoid errors on both environment.
* Disable neosnippet in web-related languages, because Emmet is more useful.
* Enable deoplete path completions from buffer directory (not current directory).
* Disable LSP diagnostics because it's buggy when used with Neomake.
* Create empty autocmds for vim-multiple-cursors to avoid warning on exit. (This almost seems to be vim-multiple-cursors issue)

wsdjeg:
@sei40kr Thanks for you info, and I just create a new PR about the better default configration. In this PR, I will change some vim options and SpaceVim layers. all language specific layers will be disabled. these should be enabled only by users themself. 

the list of layers I want to enabled is:

1. autocomplete: for auto completion.
2. checkers: for syntax checking
3. format: for code formatting
> maybe we can remove format layer, as in this layer there is only one plugin, and one key binding. all formatting option are setted in lang#* layer.
4. ui layer: SpaceVim ui, statusline and tabline should be move to this layer.
5. unite layer, I can not disable unite layer, as vimfiler requires unite.vim
6. denite: if py3 is enabled, this layer should be enabled
7. ctrlp: if no py/py3, then use this layer

maybe we can use leaderf for vim with only `+py` feature.

sei40kr:
I've had wrong idea about better default.
It's just like better-defaults layer in Spacemacs, right?
In that case, I'd prefer to enhance f/F/t/T behavior with clever-f or easymotion.
They're very useful because they have less strokes than `SPC j j`.

* https://github.com/rhysd/clever-f.vim
* https://github.com/easymotion/vim-easymotion

wsdjeg:
@sei40kr I am not sure if we should change current behaviour of `f/F`, as there have been used since SpaceVim v0.1.0.

jetm:
I like the idea of @sei40kr, also this doesn't deviate that much of Vim. People like me, coming of learning Vim will find it difficult to unlearn it.

[...]

PR #1460, opened 2018-03-04
Add clever f

wsdjeg:
ref: #1436


autoload/SpaceVim.vim
```diff
@@ -69,7 +69,7 @@ let g:spacevim_windows_leader          = 's'
 ""
 " Unite work flow leader of SpaceVim. Default is `f`.
```
jetm, line 70: Here should says: Default is `\f`.

autoload/SpaceVim.vim
```diff
@@ -69,7 +69,7 @@ let g:spacevim_windows_leader          = 's'
...
 " Set to empty to disable this feature, or you can set to another char.
-let g:spacevim_unite_leader            = 'f'
+let g:spacevim_unite_leader            = '\f'
 ""
 " Denite work flow leader of SpaceVim. Default is `F`.
```
jetm, line 74: I was expecting F will be \F as well as for unite.

jetm (changes requested):
I will make these changes before merge it

wsdjeg:
maybe we can remove this option, and denite layer is same as unite layer. also same as ctrp layer, user should use only one of them, and these there layer maybe should has same key bindings. the prefix can be setted when loadding layer.

zhigang1992:
hey guys, I'm having trouble use denite's lead key, which was `F`. Should that be mentioned somewhere and how should I fix that?

wsdjeg:
that was removed. you can load denite layer. and read denite layer documentation.

PR #1539, opened 2018-03-23
sudo write does not work in neovim

zhujinxuan:
# Problem
sudo tee % cannot work in neovim. see https://github.com/neovim/neovim/issues/1716

# Fix
By adding suda.vim https://github.com/lambdalisue/suda.vim


wsdjeg:
sorry. I do not want to add plugins to core layer now. my current work is to create a mini version of spacevim. for such a simple function. We can implement without plugins. and I do not like the idea of add `suda:` to the path. it is not same as `ssh:` `file:` `http:`. which can be accept by other app.

zhujinxuan:
Cool.  I will isolate `w!! and W` in a sudo layer and get the suda logic about `w !sudo ` into that sudo layer.  Then it would prevent adding `suda://` and new plugins

wsdjeg:
@zhujinxuan Instead of add a plugin, maybe we can just provide a command `:W`, which support `:W!` for sudo permission.

wsdjeg:
In SpaceVim, we have a command `:W`, but it does not support `:W!`, maybe we can  improve it.

zhujinxuan:
@wsdjeg What is you preferred interface for sudo read? map `r!!` or ?

zhujinxuan:
update: isolate sudo logics in a single layer.  Do not import `suda://` and suda plugin and use the suda code inside sudo.vim

wsdjeg:
@zhujinxuan I am not sure if we can do this, as suda.vim is under MIT LICENSE

zhujinxuan:
@wsdjeg  Got it. I will rewrite all these functions.

wsdjeg:
@zhangjiayin just some tips:

1. old version of vim has no function `execute()`, so you can use SpaceVim api

```vim
let s:CMP = SpaceVim#api#import('vim#compatible')
let result = s:CMP.execute(cmd)
```

2. you can move suda layer into core layer. as this feature is included in core layer by default.

3. if you defind command `W`. you can use `cnoremap w!! W`

4. we should avoid global func `SudoWriteCurrentFile`, you can defind func like this:

```vim
    call SpaceVim#mapping#space#def('nnoremap', ['f', 'W'], 'call call('
          \ . string(s:_function('s:SudoWriteCurrentFile')) . ', [])',
          \ 'save buffer with sudo', 1)

command! W call <SID>SudoWriteCurrentFile()
func! s:SudoWriteCurrentFile()
endf
```

zhujinxuan:
@wsdjeg The tips are great.  Thank you a lot.

zhujinxuan:
I will ask the author of suda.vim about whether we can re-use his code.

I have a replace in using the snippet. But vim think the `s:_function` is an undefined variable.  Using `s:_function` will cause config function abort.

I think we can add suda layer as an optional layer, and then map r!! as suda read. (Because I cannot come up wih no easy way to implement a sudo read)

wsdjeg:
```vim
" function() wrapper
if v:version > 703 || v:version == 703 && has('patch1170')
  function! s:_function(fstr) abort
    return function(a:fstr)
  endfunction
else
  function! s:_SID() abort
    return matchstr(expand('<sfile>'), '<SNR>\zs\d\+\ze__SID$')
  endfunction
  let s:_s = '<SNR>' . s:_SID() . '_'
  function! s:_function(fstr) abort
    return function(substitute(a:fstr, 's:', s:_s, 'g'))
  endfunction
endif
```

this wrap is for old vim support.

zhujinxuan:
@wsdjeg Sorry, I mean neovim thinks s:_function is an undefined variable.

wsdjeg:
you need add this wrapper into sudo.vim

zhujinxuan:
Thanks for your wrapper.  That works~

lambdalisue:
If you don't like `suda://` prefix, use `let g:suda_startup = 0` to prevent initialization. You can use `suda#write({expr}[, {options}])` or `suda#read({expr}[, {options}])` directly without initialization.

<del>I'm OK if you copied the code while you follow the MIT license</del> but I don't think it's a good idea to copy the code while you can use suda.vim for your use-case.

lambdalisue:
wait wait wait. No. This is GPL. I'm really afraid that.

wsdjeg:
@lambdalisue  yeah， that is way I need to talk with you

lambdalisue:
I think MIT license does not restrict usage of the code as long as users follow the license so I think I cannot stop you to copy my codes in GPL project but I would really want to say NO while I would not be happy for getting involved in GPL things.

I'm not a lawyer so I'm really nervous to get involved in GPL things.

lambdalisue:
If you really need to copy the code (I'm nervous but...), please make sure you follow the MIT license **properly**. You probably need to copy the entire license into the script and you probably need to section the code to classify which code has copied.

wsdjeg:
@lambdalisue  I have use many of  your plugins ，same as Shougo's，so  I need your supports. I will not copy the code.

zhujinxuan:
Thanks to @wsdjeg and @lambdalisue .  I update this PR by loading suda.vim with `let g:suda_startup = 0`

Then the only problem is how to auto-reload buffer after write.  I use a checktime here, but it would need a user confirmation.

PR #1596, opened 2018-04-03
Add vimcompatible mode

wsdjeg:


NGenetzky:
It would appear that the following idiom should be used to set defaults (rather than simple assignments):
```
let g:spacevim_windows_smartclose = get(g:, 'spacevim_windows_smartclose', 'q')
```
I believe it would be more logical to keep the compatible close to the default, such as:
```
if g:spacevim_vimcompatible == 0
  let g:spacevim_windows_smartclose = get(g:, 'spacevim_windows_smartclose', 'q')
else
  " TODO: Find suitable alternative to 'q', for now just disable.
  let g:spacevim_windows_smartclose = get(g:, 'spacevim_windows_leader', '')
endif
```

autoload/SpaceVim.vim
```diff
@@ -76,7 +76,7 @@ let g:spacevim_unite_leader            = '\f'
 ""
 " Denite work flow leader of SpaceVim. Default is `F`.
 " Set to empty to disable this feature, or you can set to another char.
-let g:spacevim_denite_leader            = 'F'
+let g:spacevim_denite_leader            = '\f'
```
NGenetzky, line 79: This does not appear to be related to 'vimcompatible' changes, but i assume that is intentional.

Don't forget to update the comment on line 77. Likewise, 73 and 75 appear to not match.

NGenetzky (changes requested):
The way you use 'origin' sounds strange. 
I would recommend updating to something like: "In vimcompatible mode attempts are made to not map over existing vim key bindings"

wsdjeg:
@NGenetzky `autoload/SpaceVim.vim` is loaded before curstom config file `~/.SpaceVim.d/init.vim`, so no need to use `let g:spacevim_windows_smartclose = get(g:, 'spacevim_windows_smartclose', 'q')`

PR #1651, opened 2018-04-22
Fix error `unknown option guifont` in VimR

zhujinxuan:
For neovim gui front-ends, the neovim is started without gui and they set gui in their own ways.  For example, starting neovim in VimR+SpaceVim throws an error that:

screen shot 2018-04-22 at 5 30 04 pm

This PR checks about if has('gui_running') before set guifont

### DIFF
--- a/config/general.vim
+++ b/config/general.vim
@@ -28,8 +28,6 @@ endif
 if g:spacevim_hiddenfileinfo == 1 && has('patch-7.4.1570')
     set shortmess+=F
 endif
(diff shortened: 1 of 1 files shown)


wsdjeg (commented):
use `if has('gui_running') && !empty(g:spacevim_guifont)` will be better.

zhujinxuan:
@wsdjeg Thanks.  The `if` is fixed.

PR #1653, opened 2018-04-23
Only Reload when the filetype is javascript

zhujinxuan:
I made a mistake when implementing the `auto_fix` feature in javascript.vim, which causes reloading in all neomake jobs.

This PR makes the auto_fix reload happens when and only when the filetype is javascript.

### DIFF
--- a/autoload/SpaceVim/layers/lang/javascript.vim
+++ b/autoload/SpaceVim/layers/lang/javascript.vim
@@ -65,9 +65,9 @@ function! SpaceVim#layers#lang#javascript#config() abort
   if s:auto_fix
     augroup SpaceVim_lang_javascript
       autocmd!
-      autocmd User NeomakeFinished checktime
+      autocmd User NeomakeFinished call <SID>checktime_if_javascript()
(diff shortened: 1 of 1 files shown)


autoload/SpaceVim/layers/lang/javascript.vim
```diff
@@ -126,4 +126,10 @@ function! s:tern_go_to_def() abort
   endif
 endfunction
 
+function! s:checktime_if_javascript() abort
+  if (match(&filetype, '^javascript') == 0)
+    checktime
```
wsdjeg, line 131: why do not use `if &ft ==# 'javascript'`

zhujinxuan: Because there are filetypes like `javascript.jsx`... (lol)

wsdjeg: but use two steps is slow, we can change it to `if &ft =~# '^javascript'`

zhujinxuan: @wsdjeg Great. I fix it in https://github.com/SpaceVim/SpaceVim/pull/1655

PR #1706, opened 2018-05-09
Add auto fix feature to typescript

zhujinxuan:
1. Add --fix feature to typescript.vim
2. Because the `autocmd User NeomakeFinished` are conflicting between javascript.vim and typescript.vim, I remove the filetype check to ensure auto fix can work on both ts and js files
3. fix indent and some other styling inconsistence


zhujinxuan:
close this PR; because tslint --fix is very unstable

wsdjeg:
OK

PR #1769, opened 2018-05-30
Fix syntax highlight conflict under flow-typed js

zhujinxuan:
When js is typed by flow, yajs cannot correctly highlight the js file.

This PR adds an option to disable yajs when flow is enable.


wsdjeg:
flow is too short, and maybe will make users feel confusing. how about `enable_flow_syntax`?

and the layer doc need to be updated.

zhujinxuan:
@wsdjeg Fixed;  I also add the `flowtype/vim-flow` in the recent commit.

wsdjeg:
Merged, thanks.

PR #1848, opened 2018-06-19
faq: using SpaceVim without overwrite vimrc

### LINKED ISSUE #1844: Feature request: don't overwrite .vimrc/init.vim
It would be really great if I could try SpaceVim without overwriting my .vimrc or init.vim. I'm not used to SpaceVim's workflow, and I can't afford to disrupt my current workflow. If SpaceVim created a file (e.g. `svim`) that invoked nvim or vim with a different configuration file, that would be much preferable to the current situation, and would allow me to learn to adapt to SpaceVim gradually.

Also, the installer *has* to warn you that your current configuration will be overwritten. It's not acceptable not to do that.

wsdjeg:
close #1844

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -9,13 +9,30 @@ this is a list of most asked questions about SpaceVim.
 
 <!-- vim-markdown-toc GFM -->
 
-- [Why use toml file as default configuration file?](#why-use-toml-file-as-default-configuration-file)
+- [Have a try with SpaceVim without overwrite vimrc?](#have-a-try-with-spacevim-without-overwrite-vimrc)
+- [Why use toml as the default configuration file format?](#why-use-toml-as-the-default-configuration-file-format)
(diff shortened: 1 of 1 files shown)


bluddy:
I think this is good, but perhaps this could also be the default behavior, both for vim and neovim? As a distribution, you don't want to overwrite the tools you're using -- they have their own 'distribution'.

wsdjeg:
@bluddy, sorry, I can not make this behavior default, It is same behavior as spacemacs.

bluddy:
But why do you need to be constrained by what spacemacs does?

wsdjeg:
@bluddy ok, if we make this behavior default, It will break many things, when using gvim, you can not load SpaceVim, gvim do not using alias. 

This is just a vim config, it do not contains a exetable file, if we include vim into this repo, I would like to use new path instead of `~/.vim`.

PR #1896, opened 2018-07-02
Add doc for SpaceVim apis

wsdjeg:
### PR Prelude

- [ ] I have updated the [Following-HADE](https://github.com/SpaceVim/SpaceVim/blob/master/wiki/en/Following-HEAD.md) page for this PR.

add doc of spacevim apis.

### DIFF
--- a/docs/api/password.md
+++ b/docs/api/password.md
@@ -1,0 +1,34 @@
+---
+title: "password api"
+description: "password API provides some besic functions for generating password"
+---
+
+# [Available APIs](../) >> password
+
+<!-- vim-markdown-toc GFM -->
+
+- [Intro](#intro)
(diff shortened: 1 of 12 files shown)


docs/api/password.md
```diff
@@ -0,0 +1,34 @@
+---
```
crisidev, line 1: Out of curiosity: what is the benefit of having a pure vim script to generate password when there are already very well tested core utils which are doing the same and just need to be shelled out. It is unlikely we will be able to reach the same level of testing projects like https://linux.die.net/man/1/pwgen have.

wsdjeg: Just because this command is not installed by default in other os.

PR #1979, opened 2018-07-26
Fixed: refactor fzf layer with 'fzf.vim'

wy-z:
Signed-off-by: weiyang <weiyang.ones@gmail.com>

### PR Prelude

https://github.com/junegunn/fzf.vim is good enough ~

### DIFF
--- a/autoload/SpaceVim/layers/fzf.vim
+++ b/autoload/SpaceVim/layers/fzf.vim
@@ -12,8 +12,9 @@ let s:CMP = SpaceVim#api#import('vim#compatible')
 function! SpaceVim#layers#fzf#plugins() abort
   let plugins = []
   call add(plugins, ['junegunn/fzf',                { 'merged' : 0}])
-  call add(plugins, ['Shougo/neoyank.vim', {'merged' : 0}])
-  call add(plugins, ['SpaceVim/fzf-neoyank',                { 'merged' : 0}])
+  call add(plugins, ['junegunn/fzf.vim',            { 'merged' : 0}])
+  call add(plugins, ['Shougo/neoyank.vim',          { 'merged' : 0}])
+  call add(plugins, ['SpaceVim/fzf-neoyank',        { 'merged' : 0}])
   return plugins
 endfunction
 
@@ -21,8 +22,10 @@ endfunction
 let s:filename = expand('<sfile>:~')
 let s:lnum = expand('<slnum>') + 2
 function! SpaceVim#layers#fzf#config() abort
+  let g:fzf_command_prefix = 'Fzf'
+
   let lnum = expand('<slnum>') + s:lnum - 1
-  call SpaceVim#mapping#space#def('nnoremap', ['b', 'b'], 'Fzfbuffers', 'List all buffers', 1)
+  call SpaceVim#mapping#space#def('nnoremap', ['b', 'b'], 'FzfBuffers', 'List all buffers', 1)
   call SpaceVim#mapping#space#def('nnoremap', ['p', 'f'],
         \ 'FzfFiles',
         \ ['find files in current project',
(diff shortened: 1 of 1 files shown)


wsdjeg:
sorry, I do not want to included fzf.vim, as it is just a wrapper of fzf, which can be done in this layer simplly, and easy for debug.

wsdjeg:
also see #1725

wy-z:
Uh...but I think https://github.com/junegunn/fzf.vim is good enough, and there are so many code were copied from it

wy-z:
If you think it is not good enough, you could try to improve it ~
https://github.com/junegunn/fzf.vim was created by `junegunn`, I think he has more time and ability to guarantee the quality of it

wsdjeg:
```
fzf in itself is not a Vim plugin, and the official repository only provides the basic wrapper function for Vim and it's up to the users to write their own Vim commands with it. However, I've learned that many users of fzf are not familiar with Vimscript and are looking for the "default" implementation of the features they can find in the alternative Vim plugins.

This repository is a bundle of fzf-based commands and mappings extracted from my .vimrc to address such needs. They are not designed to be flexible or configurable, and there's no guarantee of backward-compatibility.
```
**it's up to the users to write their own Vim commands with it.**, I think we should implement the command in SpaceVim. that will be easy to debug.

wy-z:
I like `SpaceVim` and I think it is an excellent distribution, but I think it should focus on organization but not try to warp every thing ~

wsdjeg:
If you really like to use fzf.vim, maybe you can create a private layer in your SpaceVim config. for example

`~/.SpaceVim.d/autoload/SpaceVim/layer/anotherfzf.vim`

then enable the layer in your toml file

```toml
[[layers]]
    name = 'anotherfzf'
```

wsdjeg:
SpaceVim has 5 fuzzy find layer, we need to key compatible with other 4 fuzzy finder layer, fzf.vim do not support hard config, as it says in readme `They are not designed to be flexible or configurable`.

in SpaceVim, if we can implemnet the feature, I do not want to use third_part plugins. 

why?

for example, in old version of SpaceVim, we use vim-rooter for detect the project root directory, but that plugin has some bugs, and the author do not want to fix it, But I need to use the feature. so How could I do? I reimplement a new project manager plugin.

wy-z:
Thanks, I will ~

wy-z:
And there are some problems with the current fzf layers ~

wsdjeg:
@wy-z SpaceVim provide an idea of how to manager plugins and config, dark vim user should has there own private layer, for example my record layer provides feature for recording gif in vim.

yeah,

> And there are some problems with the current fzf layers ~

yeah, that should be fixed. just same as #1725

wy-z:
- https://github.com/SpaceVim/SpaceVim/blame/master/autoload/SpaceVim/layers/fzf.vim#L36 
If I want to`jump to a definition in buffer` I have to install `denite` layer
- `FzfOutline` doesn't work well in my environment, see ~
image

wy-z:
`FzfOutline` has so many useless blank lines in output ~

wy-z:
`BTags` works well~
image

wy-z:
https://github.com/SpaceVim/SpaceVim/blame/master/autoload/SpaceVim/layers/fzf.vim#L267
Why do you rewrite these codes? https://github.com/junegunn/fzf.vim/blob/master/autoload/fzf/vim.vim#L239 works fine ~

wy-z:
@wsdjeg Thank you for your contribution on `SpaceVim` ~
I personally think https://github.com/junegunn/fzf.vim is acceptable ~ it is not the core dependency of `SpaceVim`, but a supplementary layer~
You can close the PR if you want~
Thanks again ~

wsdjeg:
Ok, thanks, :)

PR #2136, opened 2018-09-04
Fixed loading config file

michaeladler:
### PR Prelude

### DIFF
--- a/autoload/SpaceVim/custom.vim
+++ b/autoload/SpaceVim/custom.vim
@@ -84,7 +84,7 @@ function! SpaceVim#custom#apply(config) abort
   else
     let options = get(a:config, 'options', {})
     for [name, value] in items(options)
-      exe 'let g:spacevim_' . name . ' = value'
+      exe 'let g:spacevim_' . name . ' = "' . value . '"'
       unlet value
     endfor
     let layers = get(a:config, 'layers', [])


michaeladler:
Just for the record, I don't believe my changes cause vimlint to break. It appears to me that b362f23beceb33ddd56bb7eafba1aa25fe0fc14f is responsible for that, introducing the function ` SpaceVim#custom#write(force)` with an empty body, i.e. the argument  `force` is unused and that's what vimlint complains about.

wsdjeg:
Thanks For your pr. what is the issue? I can not reproduce it in current version.

michaeladler:
The line `exe 'let g:spacevim_' . name . ' = value'` expands to `let g:spacevim_<name> = value`, where `name` is the only variable in that expression. In other words, you set the value of each option to the string "value". So, if you attempt to set any options in your `init.toml`, it will silently fail because the actual value of the variable is ignored. This is a quite serious bug.

wsdjeg:
seems strange, but it works well for me. I will test it.

wsdjeg:
@michaeladler I have use the master branch everyday, but I can not reproduce this issue, can you provide any reproduce step?  for example

```toml
[options]
   filemanager = "nerdtree"
```

then open vim, run `:echo g:spacevim_filemanager`, I think you should see `nerdtree`.

wsdjeg:
I just look into this issue, with this config:

```toml
[options]
   filemanager = "nerdtree"
```

SpaceVim will run:

```vim
let [key, value] = ['filemanager', 'nerdtree']
exe 'let g:spacevim_' . name . ' = value'
```

let last line means:

```
exe 'let g:spacevim_filemanage = value'
```

which is same as

```
let [key, value] = ['filemanager', 'nerdtree']
let g:spacevim_filemanage = value
```

PR #2185, opened 2018-09-21
Improve Ruby language layer to accept a custom REPL

thedelchop:
Taking a cue from the Lua language layer, update the Ruby language layer
to support specify a custom REPL. This command expects a valid path to
an executable, so a file path would be required if the executable is not
on your $PATH.

### PR Prelude

This change is useful because it will allow Ruby developers who have a nicely configured Pry console to work with that instead of the default irb.  Simple customization that yields a nice benefit.

### DIFF
--- a/autoload/SpaceVim/layers/lang/ruby.vim
+++ b/autoload/SpaceVim/layers/lang/ruby.vim
@@ -12,10 +12,19 @@ function! SpaceVim#layers#lang#ruby#plugins() abort
         \ ]
 endfunction
 
+let s:ruby_repl_command = ''
+
 function! SpaceVim#layers#lang#ruby#config() abort
   call SpaceVim#plugins#runner#reg_runner('ruby', 'ruby %s')
   call SpaceVim#mapping#space#regesit_lang_mappings('ruby', function('s:language_specified_mappings'))
-  call SpaceVim#plugins#repl#reg('ruby', 'irb')
+  if !empty(s:ruby_repl_command)
+      call SpaceVim#plugins#repl#reg('ruby',s:ruby_repl_command)
+  else
+      call SpaceVim#plugins#repl#reg('ruby', 'irb')
+endfunction
+
+function! SpaceVim#layers#lang#ruby#set_variable(var) abort
+  let s:ruby_repl_command = get(a:var, 'repl_command', '') 
 endfunction
 
 function! s:language_specified_mappings() abort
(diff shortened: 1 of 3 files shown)


autoload/SpaceVim/layers/lang/ruby.vim
```diff
@@ -12,10 +12,15 @@ function! SpaceVim#layers#lang#ruby#plugins() abort
         \ ]
 endfunction
 
+let s:ruby_repl_command = ''
```
wsdjeg, line 15: need a function to set the repl command;

```vim
function! SpaceVim#layers#lang#ruby#set_variable(var) abort
    let s:ruby_repl_command = get(a:var, 'repl_command', '')
endfunction
```

docs/layers/lang/ruby.md
```diff
@@ -46,8 +46,23 @@ gem install rubocop
 
 ### Inferior REPL process
 
-Start a `irb` inferior REPL process with `SPC l s i`. 
+Start a `irb` inferior REPL process with `SPC l s i`. You may change the REPL command by layer option `repl_command`. For example, if you want to use `pry`, load this layer via:
 
+```vim
+call SpaceVim#layers#load('lang#ruby'
```
wsdjeg: we use toml to config SpaceVim now, so the config example should be 


```toml
[[layers]]
    name = "lang#ruby"
    repl_command = "pry"
```

docs/layers/lang/ruby.md
```diff
@@ -46,8 +46,23 @@ gem install rubocop
...
+    \ {
+    \ 'repl_command' : 'pry',
+    \ }
+```
+
+however, if the executable is not on your $PATH, then you need to specify a complete file path.
+
+```vim
```
wsdjeg: same as above, we should use toml.

thedelchop:
Ok, updates made, thanks for the feedback!

docs/layers/lang/ruby.md
```diff
@@ -46,8 +46,21 @@ gem install rubocop
 
 ### Inferior REPL process
 
-Start a `irb` inferior REPL process with `SPC l s i`. 
+Start a `irb` inferior REPL process with `SPC l s i`. You may change the REPL command by layer option `repl_command`. For example, if you want to use `pry`, load this layer via:
```
wsdjeg, line 50: the code block is toml, so it should  be `` ```toml  ``

thedelchop: 👍 fixed and re-pushed

wsdjeg (approved):

PR #2195, opened 2018-09-24
Fixed missing endif at ruby config layer

gbrlcustodio:
This PR intends to fix the missing `endif` keyword at ruby layer introduced in the latest version


wsdjeg:
Thanks.

PR #2218, opened 2018-09-29
Update erlang layer code owner

wsdjeg:
### PR Prelude

- [ ] I have updated the [Following-HEAD](https://github.com/SpaceVim/SpaceVim/blob/master/wiki/en/Following-HEAD.md) page for this PR.

improve erlang support


ray2011:
Some problems when using erlang layer:
1. Both neomake and vim-erlang-compile does erlang syntax check 
 You can disable neomake syntax check in SpaceVim options: `lint_on_save = false` or disable vim-erlang-compile plugin.

2. Errors like `can't find head files` or `undefined parse transform 'lager_transform'`
I only solved this in the vim-erlang-compile plugin, the escript [erlang_check.erl](https://github.com/vim-erlang/vim-erlang-compiler/blob/master/compiler/erlang_check.erl) support rebar project, rebar3 project, and project with a makefile, and I work in a rebar3 project.
The escript probably has some problems in the rebar3 project:
- not support cover_opts in rebar.config
- not support find ebin directory in `checkouts_dir` defined in rebar.config
- only support rebar.config in project dir, not support rebar.config in application dir

3. The vim-erlang-compile plugin is a little slow because it uses a slow rebar3 command `QUIET=1 rebar3 as 'profile' path` to find ebin dirs in the project.

4. The vim-erlang-omnicomplete plugin probably can't omnicomplete custom modules, because its default basedir passed to [erlang_complete.erl](https://github.com/vim-erlang/vim-erlang-omnicomplete/blob/master/autoload/erlang_complete.erl) is you source file dir opened, not your project dir.  And it is a little slow when first generate suggestions for custom modules.

That's all my problems with the lang#erlang layer, and you can modify escript or vimscript a little to solve these problems.

wsdjeg:
1. neomake is a lint framework, and It is builtin checker for erlang, maybe we can disable it.  neomake support makeprg, and vim-erlang-compiler will generate the makeprg for current file.

2. if any works need to be done in vim script, I may help with it, but I do not familier with erlang.

3. it is slow, so when use `:make` command, it will block vim, so let use neomake, it run command asynchronously.

4. I think it can be inproved via deoplete source, instead of omni function, the deoplete source is run asynchronously.

hcs42:
@wsdjeg asked me to take a look, so here are my comments.

(2) @suzuiyuegjy is correct in all points.

* Adding cover_opts support is probably easy but there might be some issue that I don't foresee.
* The `_checkouts` dir can be added easily to vim-erlang-compiler (actually, it has been added to vim-erlang-omnicomplete). I can take care of this.
* The project dir+application dir rebar.config is not trivial to solve (the two rebar.config files need to be merged together).

(4) vim-erlang-omnicomplete does support custom modules (it will find out the project dir from the base dir).

Regarding slowness: for me it doesn't take much more than a second or two when the plugins do a syntax check or load a module's autocomplete list, but it is true that this lag is noticable, and the user experience would be better if the lag wasn't there.

PR #2356, opened 2018-12-18
Add more useful key bindings for typescript layer

ksamborski:
I've been working with typescript lately and I've noticed that only these options are available in my SpaceVim (I'm on vim 8.1) in the language layer:
- [d] show document
- [e] rename symbol

I looked at the code and there are defined more of them but only for nvim. For vim the __tsuquyomi__ plugin is used but it has many more useful features than those two listed.

This PR adds following options:
- [c] type definition
- [i] import
- [m] interface implementations
- [r] references


wsdjeg:
hello, Thank you for your pr, and I hope the two sections vim and neovim(`has('nvim')`) provide same key bindings. and more layer doc is needed. I will check this pr.

PR #2415, opened 2019-01-05
Add: coc.nvim integration enhacements

chenlijun99:
### PR Prelude

Currently, if we use `coc.nvim` as autocomplete method and enable the lsp layer, we have two Language  server clients enabled. This is quite undesirable.
Users are told to disable the lsp layer if they use coc.nvim, but this is not mentioned in the documentation.
Therefore, the lsp layer should use coc.nvim as client if it's used as autocomplete method and the lsp layer's configuration should be passed to coc.nvim through the `coc#config()` function. This way we have a single interface for managing language servers, the user no more needs to add lsp configuration to the `coc-settings.json`, which BTW has to be placed in the `~/.vim` directory and since `~/.vim` for SpaceVim users is just a link to `~/.SpaceVim`, it's relatively tricky to add `coc-settings.json` to their, let's say, dotfiles repository without polluting the SpaceVim repository.

**NOTE**
In order for this pull request to work, we have to wait neoclide/coc.nvim#316 to be merged.
Without the `CocSetup` autocmd, when `SpaceVim#lsp#reg_server` is called `coc.nvim` is not loaded yet, therefore calls to `coc#config` are invalid.

As pointed out by @chemzqm we could leverage the plugin managers' hook. I know that `vim-plug` fires an `doautocmd User <plugin>` when it loads a plugin, whereas `dein` requires the user to explicitly define this hook among the parameters passed for the plugin.

From documentation of dein
```vim
	call dein#add('Shougo/deoplete.nvim', {
	\ 'hook_source':
	\  'execute "doautocmd <nomodeline> User" "dein#source#".
	\   g:dein#plugin.name'
	\ })

```

BTW, from what I've seen, currently additional parameters for a plugin (like lazy load events or commands) are directly forwarded to the plugin manager being used. 
https://github.com/SpaceVim/SpaceVim/blob/b7be390f57c1d714483f09b238f33c24100faee7/autoload/zvim/plug.vim#L158-L192
Clearly, each one expects different keys, but I didn't find any adapter layer/function that converts the options for the various supported plugin manager. It seems that the passed options are recognized only by `dein`.
Do we really support other plugin manager?

### DIFF
--- a/autoload/SpaceVim/lsp.vim
+++ b/autoload/SpaceVim/lsp.vim
@@ -8,7 +8,69 @@
 
 scriptencoding utf-8
 
-if has('nvim')
+
+if SpaceVim#layers#isLoaded("autocomplete") && get(g:, "spacevim_autocomplete_method") ==# 'coc'
+  " use coc.nvim
+  let s:coc_language_servers = {}
+  let s:coc_language_servers_key_id_map = {}
+  function! SpaceVim#lsp#reg_server(ft, cmds) abort
+    " coc.nvim doesn't support key values containing dots
+    " See https://github.com/neoclide/coc.nvim/issues/323
+    " Since a:cmds[0], i.e. the language server command can be a full path,
+    " which can potentially contain dots, we just take it's last part, if any
+    " dots are present.
+    " 
+    " Clearly, with this implementation, an edge case could be the following
+    "
+    " [layers.override_cmd]
+    "   c = ['/home/user/.local/.bin/ccls', '--log-file=/tmp/ccls.log']
+    "   cpp = ['/home/user/local/.bin/ccls', '--log-file=/tmp/ccls.log']
+    "
+    " the last part `bin/ccls` is the same, whereas the commands are not
+    " actually the same.
+    " We need to keep an id to distinguish among conflicting keys.
+    
+    if stridx(a:cmds[0], ".") >= 0 
+      let l:key = split(a:cmds[0], "\\.")[-1]
+    else
+      let l:key = a:cmds[0]
+    endif
+
+    for id in range(get(s:coc_language_servers_key_id_map, l:key, 0))
+      if has_key(s:coc_language_servers, l:key . id) && s:coc_language_servers[l:key . id].command ==# a:cmds[0]
+        call add(s:coc_language_servers[l:key . id].filetypes, a:ft)
+        return
+      endif
+    endfor
+
+    let s:coc_language_servers_key_id_map[l:key] = get(s:coc_language_servers_key_id_map, l:key, 0)
+    let s:coc_language_servers[l:key . s:coc_language_servers_key_id_map[l:key]] = {
+          \'command': a:cmds[0],
+          \'args': a:cmds[1:],
+          \'filetypes': [a:ft]
+          \}
+
+    let s:coc_language_servers_key_id_map[l:key] = s:coc_language_servers_key_id_map[l:key] + 1
+
+    autocmd! User CocNvimInit :call coc#config("languageserver", s:coc_language_servers)
+  endfunction
+
+  function! SpaceVim#lsp#show_doc() abort
+    call CocActionAsync("doHover")
+  endfunction
+
+  function! SpaceVim#lsp#go_to_def() abort
+    call CocActionAsync("jumpDefinition")
+  endfunction
+
+  function! SpaceVim#lsp#rename() abort
+    call CocActionAsync("rename")
+  endfunction
+
+  function! SpaceVim#lsp#references() abort
+    call CocActionAsync("jumpReferences")
+  endfunction
+elseif has('nvim')
   " use LanguageClient-neovim
   function! SpaceVim#lsp#reg_server(ft, cmds) abort
     let g:LanguageClient_serverCommands[a:ft] = copy(a:cmds)
--- a/autoload/SpaceVim/layers/lsp.vim
+++ b/autoload/SpaceVim/layers/lsp.vim
@@ -9,7 +9,9 @@
 function! SpaceVim#layers#lsp#plugins() abort
   let plugins = []
 
-  if has('nvim')
+  if SpaceVim#layers#isLoaded("autocomplete") && get(g:, "spacevim_autocomplete_method") ==# 'coc'
+    " nop
+  elseif has('nvim')
     call add(plugins, ['autozimu/LanguageClient-neovim',
           \ { 'merged': 0, 'if': has('python3'), 'build' : 'bash install.sh' }])
   else
--- a/docs/layers/autocomplete.md
+++ b/docs/layers/autocomplete.md
@@ -10,6 +10,7 @@ description: "Autocomplete code within SpaceVim, fuzzy find the candidates from 
 - [Description](#description)
 - [Install](#install)
 - [Configuration](#configuration)
+  - [Choose which completion engine to be used](#choose-which-completion-engine-to-be-used)
   - [Key bindings](#key-bindings)
   - [Snippets directories](#snippets-directories)
   - [Show snippets in auto-completion popup](#show-snippets-in-auto-completion-popup)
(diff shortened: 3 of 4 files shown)


wsdjeg:
I will implement a compatible function for different plugin manager.

autoload/SpaceVim/lsp.vim
```diff
@@ -8,7 +8,38 @@
 
 scriptencoding utf-8
 
-if has('nvim')
+if g:spacevim_autocomplete_method ==# 'coc'
+  " use coc.nvim
+  let s:coc_language_servers = {}
+  function! SpaceVim#lsp#reg_server(ft, cmds) abort
```
wsdjeg, line 16: 1. the argument `cmds` maybe just a string. 
2. if the cmds is a list, the first argv maybe the whole path of the executable

chemzqm:
Support for change of languageserver configuration after server start is added, no need to use `CocSetup`

wsdjeg:
Just read the config of coc, why it use language server as key in the dict? I think use filetype will be better, because if two server has same filetype, It will cause error.

________________________________
wsdjeg@outlook.com

autoload/SpaceVim/lsp.vim
chenlijun99: As to the first point, I think the the documentation doesn't mention the possibility of cmds being a string.
Furthermore, I found this function which also assumes `cmds` to be an array

https://github.com/SpaceVim/SpaceVim/blob/9634851a1dc6be2b2aaa124c105ccc010a9b70fc/autoload/SpaceVim/layers/lsp.vim#L103-L120

chenlijun99: As to the second point, please see comments below

chenlijun99:
As @wsdjeg said, there is possibility that `a:cmds[0]` is a full path executable.
I first thought that it shouldn't be a problems, but then I encountered a corner case issue with `coc.nvim`'s configuration. Please see neoclide/coc.nvim#323.

If @chemzqm thinks that this is a "won't fix", there is a simple solution:
Use as key only the final part the path. But there is a problem. Let's say I have this configuration
```
[[layers]]
  name = 'lsp'
  filetypes = [
    'c',
    'cpp'
    'typescript'
  ]
  [layers.override_cmd]
    c = ['/home/user/.local/bin/ccls']
    cpp = ['ccls'']
```
One wants to use the local LSP for C and the LSP in $PATH for C++.
With this approach, we will end up using one LSP (the first one that is saved in the dict) for both filetypes.

chenlijun99:
Since neoclide/coc.nvim#323 has been resolved with a sort of "won't fix" (which is what I expected, since AFAICT most Javascript "get by key" functions don't support a key containing a dot), a solution has to be found at SpaceVim's side.
Extending the solution that I proposed in https://github.com/SpaceVim/SpaceVim/pull/2415#issuecomment-451732716 (which in short is: in case the command is a full path, just take the final part), I think we can check whether a key does already exist and if yes check whether the command property of the value associated with that key is the same. If they're different, it means that we have to add a new item to the list, whose key can be the original key + a self increment counter.

wsdjeg:
https://github.com/neoclide/coc.nvim/issues/323#issuecomment-451796261

My be we should not use command as the server id.

autoload/SpaceVim/lsp.vim
```diff
@@ -8,7 +8,38 @@
...
+    if has_key(s:coc_language_servers, a:cmds[0])
+      call add(s:coc_language_servers[a:cmds[0]].filetypes, a:ft)
+    else
+      let s:coc_language_servers[a:cmds[0]] = {
+            \'command': a:cmds[0],
+            \'args': a:cmds[1:],
+            \'filetypes': [a:ft]
+            \}
+    endif
+    autocmd! User CocSetup :call coc#config("languageserver", s:coc_language_servers)
```
wsdjeg: `SpaceVim#lsp#reg_server()` will be called in every language layer, when this function is called, an autocmd is definded, should we move this autocmd to coc's config.
```
config/plugins/coc.vim
```

chenlijun99: Note that with the `!` at the end of the `autocmd`, the `autocmd` will be removed and re-registered at each execution. So at the end we have only one `autocmd` registered. Still, this may lead to some overhead, but IMHO it's quite unnoticeable.

However, if you think that it's preferable to call `coc#config` in `config/plugins/coc.vim`, I'm totally fine with it.
In that case, do you know a way to share `s:coc_language_servers` with `config/plugins/coc.vim`? I'm not familiar with Vimscript and on Google I didn't find anything.
A quick solution would be to make `coc_language_servers` global (so `g:coc_language_servers`), so as to make it accessible to `config/plugins/coc.vim`, but I don't know whether it's fine for you to pollute the global namespace.

wsdjeg:
Ok, I am ok with this PR now. can we merge it?
 I do not want to use global namespace, as It will inscrease the completion items in cmdline.

________________________________
wsdjeg@outlook.com

chenlijun99:
Now the use of full path as lsp commands should be fully supported.

Given a configuration like this
```toml
  [layers.override_cmd]
    c = ['/home/lijun/.local/.bin/ccls', '--log-file=/tmp/ccls.log']
    cpp = ['/home/lijun/local/.bin/ccls', '--log-file=/tmp/ccls.log']
```

the following configuration is passed to coc with `coc#config`
```json
{ 
	"bin/ccls0": {
		"args": ["--log-file=/tmp/ccls.log"],
		"filetypes": ["c"],
		"command": "/home/lijun/.local/.bin/ccls"
	},
	"bin/ccls1": {
		"args": ["--log-file=/tmp/ccls.log"],
		"filetypes": ["cpp"],
		"command": "/home/lijun/local/.bin/ccls"
	}
} 
```

chenlijun99:
If for you it's OK now, I think we can merge this PR.

wsdjeg:
Ok， this branch is out-of-date, so I squash and merge it.

Maxattax97:
This appears to no longer work due to this commit, about 2 months ago.

https://github.com/neoclide/coc.nvim/commit/8639687823ac44eb573af3b014b14116d1e223b4#diff-47aba0262aa96b8f4b3de5f46b872d4dR182

Because of this, I am now seeing errors like this:
```
[coc.nvim] Invalid languageserver id: docker-langserver0, only character allowed!                                                                                           
[coc.nvim] Invalid languageserver id: bash-language-server0, only character allowed!                                                                                        
[coc.nvim] Invalid languageserver id: javascript-typescript-stdio0, only character allowed!                                                                                 
[coc.nvim] Invalid languageserver id: eclipse-jdt0, only character allowed!   
```

For configuration like this:
```toml
[[layers]]
    name = "lsp"
    filetypes = [
        "sh", # Install: npm install -g bash-language-server
        "dockerfile", # Install: npm install -g dockerfile-language-server-nodejs
        "javascript", # Install: npm install -g javascript-typescript-langserver
        "c", # Install: cquery (https://github.com/cquery-project)
        "cpp",
        "java" # Install: eclipse-jdt (https://github.com/eclipse/eclipse.jdt.ls)
    ]
    [layers.override_cmd]
        dockerfile = ["docker-langserver", "--stdio"]
        c = ["cquery", "--log-file=/tmp/cquery.log", '--init={"cacheDirectory":"/tmp/cquery/", "completion": {"filterAndSort": false}}']
        cpp = ["cquery", "--log-file=/tmp/cquery.log", '--init={"cacheDirectory":"/tmp/cquery/", "completion": {"filterAndSort": false}}']
        java = ["eclipse-jdt"]
```

chenlijun99:
Hi,
just like the problem this pull request should have addressed, again the bug that we're facing now can be either solved at coc.nvim side or at SpaceVim side.

IMHO it would be easier to ask @chemzqm to change the behaviour at coc.nvim's side, since it's pretty normal to have names with hyphens and actually the initial problem was only caused by the dot character.

wsdjeg:
well, I can do nothing with it, I think the plugin ( Whether COC or vim-lsp or even language-client.nvim) should provide a similar configuration interfaces.

Maxattax97:
I agree with @free-easy. I've started the discussion over on [COC's repo](https://github.com/neoclide/coc.nvim/issues/502).

Maxattax97:
This issue has been resolved.
https://github.com/neoclide/coc.nvim/commit/12e697ba040fffd179d7e48b435dc2c1a807ac10#diff-47aba0262aa96b8f4b3de5f46b872d4dL181

PR #2440, opened 2019-01-08
Enhance c/c++ layer

chenlijun99:
### PR Prelude

### DIFF
--- a/autoload/SpaceVim/layers/lang/c.vim
+++ b/autoload/SpaceVim/layers/lang/c.vim
@@ -88,6 +88,10 @@ function! SpaceVim#layers#lang#c#plugins() abort
 endfunction
 
 function! SpaceVim#layers#lang#c#config() abort
+  call SpaceVim#mapping#gd#add('c',
+        \ function('s:go_to_def'))
+  call SpaceVim#mapping#gd#add('cpp',
+        \ function('s:go_to_def'))
   call SpaceVim#plugins#runner#reg_runner('c', ['gcc -o #TEMP# %s', '#TEMP#'])
   call SpaceVim#mapping#space#regesit_lang_mappings('c', function('s:language_specified_mappings'))
   call SpaceVim#plugins#runner#reg_runner('cpp', ['g++ -o #TEMP# %s', '#TEMP#'])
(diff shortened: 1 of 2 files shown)


autoload/SpaceVim/layers/lang/c.vim
```diff
@@ -88,6 +88,10 @@ function! SpaceVim#layers#lang#c#plugins() abort
 endfunction
 
 function! SpaceVim#layers#lang#c#config() abort
+  call SpaceVim#mapping#gd#add('c',
+        \ function('s:go_to_def'))
```
wsdjeg, line 92: we should defind two function, here is a example, user who enable lsp for c but not for cpp. he should be able to use lsp gd

autoload/SpaceVim/layers/lang/c.vim
```diff
@@ -194,3 +200,10 @@ function! s:update_neoinclude(argv, fts) abort
   let b:neoinclude_paths = path
 endfunction
 
+function! s:go_to_def() abort
+  if !SpaceVim#layers#lsp#check_filetype('c') && !SpaceVim#layers#lsp#check_filetype('cpp')
```
wsdjeg: If you do not want to use two function, you can change this line to:

```
if !SpaceVim#layers#lsp#check_filetype(&ft)
```

chenlijun99:
Fixed

wsdjeg:
as said in https://spacevim.org/development/#language-specified-key-bindings

we should unify the language specified key bindings for different language layer.

chenlijun99:
Yeah, but it's not mentioned which key binding to use for "show references".
I used "SPC l \*", because vim's default behavior with "*" is "search for same word under the cursor in the buffer"

wsdjeg:
`*` is hard to press, you need to press `Shift+8`. in typescript layer, we are using `SPC l r`, but I think `SPC l r` is not a good idea, it should be used to run file.

chenlijun99:
Uh, I didn't take that into consideration. Yeah, Vim's philosophy is "stay on the home row".

What do you think about "SPC l f", where `f` stands for "find"?

wsdjeg:
`SPC l f` is ok to me. :)

________________________________
wsdjeg@outlook.com

chenlijun99:
Done

Issue #2479, opened 2019-01-21
Feature request: open terminal at current file's directory


chenlijun99:
I rarely use terminal, but when I use it often I need to operate on the current file, instead of files in the current project.

It would awesome to have something like "SPC \"" for opening a terminal in the current file's directory.

wsdjeg:
yeah, it is useful

PR #2480, opened 2019-01-21
Add: keybinding to open terminal at current file's directory

chenlijun99:
### PR Prelude

See issue #2479

### DIFF
--- a/autoload/SpaceVim/layers/shell.vim
+++ b/autoload/SpaceVim/layers/shell.vim
@@ -37,10 +37,19 @@ let s:file = expand('<sfile>:~')
 let s:lnum = expand('<slnum>') + 2
 function! SpaceVim#layers#shell#config() abort
   call SpaceVim#mapping#space#def('nnoremap', ["'"], 'call call('
-        \ . string(function('s:open_default_shell')) . ', [])',
+        \ . string(function('s:open_default_shell')) . ', [0])',
         \ ['open shell',
         \ [
         \ "[SPC '] is to open or jump to default shell window",
+        \ '',
+        \ 'Definition: ' . s:file . ':' . s:lnum,
+        \ ]
+        \ ], 1)
+  call SpaceVim#mapping#space#def('nnoremap', ["\""], 'call call('
+        \ . string(function('s:open_default_shell')) . ', [1])',
+        \ ["open shell in current file's path",
+        \ [
+        \ "[SPC \"] is to open or jump to default shell window with the current file's pwd",
         \ '',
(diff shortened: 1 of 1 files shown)


chenlijun99:
This pull request doesn't yet cover the case where a terminal is not closed but the keybinding is invoked again.

Discussions are welcome.

General ideas on how this can be addressed:

* Don't use lcd approach. Instead, on each "SPC \"", send keys to the terminal with vim's `term_sendkeys()` and neovim's `chansend()`. Something like `cd <path>\<cr>\<c-l>`.
* When switching to terminal, check whether it's working directory is the desired one. If not, close it and proceed with creating a new terminal with the desired working directory.

wsdjeg:
well, I think `SPC '` is used to open project shell, this shell can be hidden, and the history in this shell should not be cleared when hide the shell windows.

the `SPC "` key binding should be able to open new shell if no shell windows exist, the dir of the shell is buffer dir. and if the project shell is openen, `SPC "` should be same as `SPC '` which is just switch to shell windows. because for user who windows is samll, I do not think open two shell windows is a good idea.

send key is not a good idea, bucase, you do not know if the command in current current shell  widnow is exit. for example if you are run `jekylll server start` and this command is not exit, then you can not send `cd path` to that shell.

wsdjeg:
BTW, no need to use `lcd`, just start a shell windows via `term_open()`, this function support specific the cwd.

chenlijun99:
I totally overlooked the `cwd` options. That would simplify a lot of things.

Yeah, I totally agreen that send keys would be a really fragile implementation.

> the `SPC "` key binding should be able to open new shell if no shell windows exist, the dir of the shell is buffer dir. and if the project shell is openen, `SPC "` should be same as `SPC '` which is just switch to shell windows. because for user who windows is samll, I do not think open two shell windows is a good idea.

This would be the simplest, but not the most user-friendly solution.

A harder but more user friendly solution would be to create a terminal for each different directory and to show only the active one.

wsdjeg:
> A harder but more user friendly solution would be to create a terminal for each different directory and to show only the active one.

when press `SPC "`, check if the shell windows exits, if the shell exist, check the shell dir ( which can be set when create a shell windows, use b:var), if the b:var is not same, just hide the shell windows, create a new shell for current buffer dir.

chenlijun99:
Yeah, that's exactly what I was planning to do.
Let's see if I have some time to implement it.

chenlijun99:
The corner case is now handled pretty well.

PR #2534, opened 2019-01-31
Fixed: ChineseLinter.vim should not be loaded for all file types.

xylinkun:
### PR Prelude

ChineseLinter.vim should only be load for a limit number  of filetypes, such as markdown, text."

### DIFF
--- a/autoload/SpaceVim/layers/chinese.vim
+++ b/autoload/SpaceVim/layers/chinese.vim
@@ -10,8 +10,8 @@
 function! SpaceVim#layers#chinese#plugins() abort
   let plugins = [
         \ ['yianwillis/vimcdoc'          , {'merged' : 0}],
-        \ ['ianva/vim-youdao-translater' , {'on_cmd' : ['Ydv' , 'Ydc', 'Yde']}],
-        \ ['wsdjeg/ChineseLinter.vim'    , {'merged' : 0, 'on_cmd' : ['CheckChinese']}],
+        \ ['ianva/vim-youdao-translater' , {'merged' : 0, 'on_cmd' : ['Ydv' , 'Ydc', 'Yde']}],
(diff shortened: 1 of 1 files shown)


autoload/SpaceVim/layers/chinese.vim
```diff
@@ -10,8 +10,8 @@
 function! SpaceVim#layers#chinese#plugins() abort
   let plugins = [
         \ ['yianwillis/vimcdoc'          , {'merged' : 0}],
-        \ ['ianva/vim-youdao-translater' , {'on_cmd' : ['Ydv' , 'Ydc', 'Yde']}],
-        \ ['wsdjeg/ChineseLinter.vim'    , {'merged' : 0, 'on_cmd' : ['CheckChinese']}],
+        \ ['ianva/vim-youdao-translater' , {'merged' : 0, 'on_cmd' : ['Ydv' , 'Ydc', 'Yde']}],
```
wsdjeg, line 13: lazy loaded plugin is unmerged by default, that means, if you use `on_cmd`, no need to add `merged` key.

autoload/SpaceVim/layers/chinese.vim
```diff
@@ -10,8 +10,8 @@
...
+        \ ['wsdjeg/ChineseLinter.vim'    , {'merged' : 0, 'on_cmd' : 'CheckChinese', 'on_ft' : ['markdown', 'text']}],
```
wsdjeg, line 14: This plugin is not runtime plugin, It is a command plugin, so no need to use on_ft feature, just use on_cmd is ok.

xylinkun: Is it OK that Checkchinese command key binding appears in all file type?

PR #2655, opened 2019-03-11
Add: runnning command for compiling languages can be determined by language layers

SleepyBag:
…language layers now.

### PR Prelude

For now in runner.vim, when a runner of compiling language is being processed in async_run, the running command is totally  controlled by runner.vim, who just generates a random filename and ignores the second runner object passed by parameter. 
Consequently, language layers are only able to influence the compiling command, but not the running one. 
However, sometimes users need to attach some parameters to the runnning command. For example, I'd like to redirect the input stream of my cpp program to a pre-wrote file, instead of stdin, by attaching " < input.txt" to the running command.
After this modification, compiling language layers can control the running command by the second runner object, which is actually a string. For instance, when this object is "#TEMP# < input.txt", #TEMP# will be replaced by the actual temporary filename and "filename < input.txt" will be run.


autoload/SpaceVim/plugins/runner.vim
```diff
@@ -71,26 +71,29 @@ function! s:async_run(runner) abort
   elseif type(a:runner) == type([])
     " the runner is a list
     " the first item is compile cmd, and the second one is running cmd.
-    let s:target = s:FILE.unify_path(tempname(), ':p')
+    let target = s:FILE.unify_path(tempname(), ':p')
+    " concatenate compiling paramaters in runner[0] into a command
     if type(a:runner[0]) == type({})
-      if type(a:runner[0].exe) == 2
+      if type(a:runner[0].exe) == type(function("tr"))
```
wsdjeg, line 97: the value of `type(function("tr"))` is 2.

wsdjeg (changes requested):

PR #2687, opened 2019-03-20
Add: option for list number of files on home

zeroryuki:
option to show list number of files in home
default is 6
ex: home_files_number = 10


rachmadaniHaryono:
why `SpaceVim-options-home_files_number` on `max_column` section?

PR #2702, opened 2019-03-26
Update denite to use file/rec instead of file_rec

stfl:

denite complains when using <kbd>SPC f f</kbd> that we should use file/rec instead of file_rec
so here is a simple search/replace PR..


wsdjeg (approved):

wsdjeg:
thanks.

PR #2825, opened 2019-05-14
Fix: disable plugin

alanding1989:
### PR Prelude

right now if I disable some plugin in `g:spacevim_disabled_plugins`, when I open denite to check plugin list, the disabled plugins also shown in the list. it has not been removed from
```
g:unite_source_menu_menus=AddedPlugins=command.candidates
```


autoload/SpaceVim/plugins.vim
```diff
@@ -21,16 +20,18 @@ function! s:load_plugins() abort
   for group in SpaceVim#layers#get()
     let g:_spacevim_plugin_layer = group
     for plugin in s:getLayerPlugins(group)
-      if len(plugin) == 2
-        call zvim#plug#add(plugin[0], plugin[1])
-        if zvim#plug#tap(split(plugin[0], '/')[-1]) && get(plugin[1], 'loadconf', 0 )
-          call zvim#plug#defind_hooks(split(plugin[0], '/')[-1])
+      if index(g:spacevim_disabled_plugins, split(plugin[0], '/')[-1]) == -1
+        if len(plugin) == 2
```
wsdjeg, line 24: you should check if len(plugin) == 2 first, otherwise plugin[0] maybe cause error.

alanding1989: I have changed the commit, pls check.

wsdjeg:
Sorry, I have checked this PR, we can not use `if index(g:spacevim_disabled_plugins, split(plugin[0], '/')[-1]) == -1` use this expr to check if the plugin is disabled, because all items in `g:spacevim_disabled_plugins` should be the key in `dein#get()`. In current version of SpaceVim, many plugins are bundle plugin, so we need to change the logic of plugin list.

PR #2868, opened 2019-06-02
Add: custom register language specific mapping function

alanding1989:
### PR Prelude

right now custom can use
```
SpaceVim#mapping#space#def() or SpaceVim#custom#SPC()
```

 to **add new** or **override default** layer mappings, but I found that it is impossible to change a default language specific mapping, if use the above function with 
```
SpaceVim#mapping#space#regesit_lang_mappings()
``` 
 it will override **all default** lang mappings when vim enter into specific filetype buffer.
in this commit add a function
```
SpaceVim#custom#Reg_langSPC()
```
custom can call it in bootstrap#**after** -- **VimEnter** with Custom-defined mappings, it will add new **undefined(i.g. customized mapping that haven`t been defined by SpaceVim)** mappings or override default **same** mappings (when the mapping key has been defined by SpaceVim), e.g.

```
function MySpaceVim#bootstrap#after()
    call SpaceVim#custom#Reg_langSPC('vim', function('s:language_specified_mappings'))

endfunction

function! s:language_specified_mappings() abort
  if g:is_spacevim
    call SpaceVim#mapping#space#langSPC('nnoremap', ['l', 't'],
          \ 'call setline(line("$"), "\" vim:set sw=2 ts=2 sts=2 et tw=78 fmd=marker")',
          \ 'insert Vim file tail', 1)
    call SpaceVim#mapping#space#langSPC('nnoremap', ['l', 'a'], 'call call('
          \ . string(s:_function('s:addParam')) . ', [])',
          \ '@ add debug Parameter', 1)
```

### DIFF
--- a/autoload/SpaceVim/custom.vim
+++ b/autoload/SpaceVim/custom.vim
@@ -83,6 +83,9 @@ function! SpaceVim#custom#SPCGroupName(keys, name) abort
   call add(g:_spacevim_mappings_space_custom_group_name, [a:keys, a:name])
 endfunction
 
+function! SpaceVim#custom#Reg_langSPC(ft, func) abort
+  call SpaceVim#mapping#space#extend_CustomLSPC({a:ft.'custom' : a:func})
+endfunction
 
 function! SpaceVim#custom#apply(config, type) abort
   " the type can be local or global
--- a/autoload/SpaceVim/mapping/space.vim
+++ b/autoload/SpaceVim/mapping/space.vim
@@ -600,14 +600,24 @@ function! s:windows_layout_toggle() abort
 endfunction
 
 
+function! SpaceVim#mapping#space#extend_CustomLSPC(dict) abort
+  call extend(s:language_specified_mappings, a:dict)
+endfunction
+
+
 let s:language_specified_mappings = {}
 function! SpaceVim#mapping#space#refrashLSPC() abort
   let g:_spacevim_mappings_space.l = {'name' : '+Language Specified'}
-  if !empty(&filetype) && has_key(s:language_specified_mappings, &filetype)
-    call call(s:language_specified_mappings[&filetype], [])
-    let b:spacevim_lang_specified_mappings = g:_spacevim_mappings_space.l
+  if !empty(&filetype)
+    if has_key(s:language_specified_mappings, &filetype)
+      call call(s:language_specified_mappings[&filetype], [])
+      let b:spacevim_lang_specified_mappings = g:_spacevim_mappings_space.l
+    endif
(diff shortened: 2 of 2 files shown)


wsdjeg:
use new filetype is not a good idea, the function will be called twice, I think if user add custom language specific function, the old one should not be called.

wsdjeg:
btw, there should be two function, one is to override the old function, one is to extend new key bindings.

alanding1989:
> use new filetype is not a good idea, the function will be called twice, I think if user add custom language specific function, the old one should not be called.

why you say use a new filetype?  **&filetype. 'custom'**  is just a key to specify the **dict-key** to the custom`s function, in order to **distinguish from** the default mapping definition function.
```
    if has_key(s:language_specified_mappings, &filetype.'custom')
      call call(s:language_specified_mappings[&filetype.'custom'], [])
      call extend(b:spacevim_lang_specified_mappings, g:_spacevim_mappings_space.l)
    endif
```
---
Your default defined mapping dict 
`let b:spacevim_lang_specified_mappings = g:_spacevim_mappings_space.l` is **generated** after event BufEnter or FileType **by calling** this function https://github.com/SpaceVim/SpaceVim/blob/49c92405123cdb48a8dcb22bb1859b600bef96b8/autoload/SpaceVim/autocmds.vim#L54

**In this commit**, this function is also called by event BufEnter, and FileType, **first** generate the default mapping defined in the source code, **then** generate the custom defined mapping and extend to the dict 
` b:spacevim_lang_specified_mappings`.  It haven`t been called twice. 

https://github.com/alanding1989/SpaceVim/blob/7ca5128a4d61cb672d6eeef4f97ca1077cb6a90e/autoload/SpaceVim/mapping/space.vim#L610
```
function! SpaceVim#mapping#space#refrashLSPC() abort
  let g:_spacevim_mappings_space.l = {'name' : '+Language Specified'}
  if !empty(&filetype)
    if has_key(s:language_specified_mappings, &filetype)
      call call(s:language_specified_mappings[&filetype], [])
      let b:spacevim_lang_specified_mappings = g:_spacevim_mappings_space.l
    endif
    if has_key(s:language_specified_mappings, &filetype.'custom')
      call call(s:language_specified_mappings[&filetype.'custom'], [])
      call extend(b:spacevim_lang_specified_mappings, g:_spacevim_mappings_space.l)
    endif
  endif
endfunction
```

wsdjeg:
```
call call(s:language_specified_mappings[&filetype], [])
```

```
call call(s:language_specified_mappings[&filetype.'custom'], [])
```

will be called at same time.

wsdjeg:
done in #3260

PR #3042, opened 2019-09-10
urlescape spaces in font names

### LINKED ISSUE #2532: install.sh: Fail to download fonts with space in file name w/o any error indications
As is in the title, in the **install.sh** provided on the Quick Start page, the bash function `download_font` relies on *curl* to downloads font from *raw.github...*:

https://github.com/SpaceVim/SpaceVim/blob/56e39b40e250d1d5e6216d11d10a1b43ba866875/docs/install.sh#L347

. But there are spaces in more than half of the bundled fonts as is listed in `install_font`, such as:

https://github.com/SpaceVim/SpaceVim/blob/56e39b40e250d1d5e6216d11d10a1b43ba866875/docs/install.sh#L359

. *curl* will get nothing other than an *HTTP 400 Bad Request* error with an empty body for all those of fonts with spaces in file name. What's worse, due to the option **-s** set, curl will silently ignore this error and continue. This leads to failing to install more than half of those fonts. I notice that users have been complaining about this issue, see #721, #771, etc (although some of these may be because of forgetting to set terminal font). 

This problem may be fixed by simply renaming those files or replace spaces by `%20` so that *curl* correctly interprets it (I am not sure how to properly escape URL in shell). Besides, to make sure no error is ignored silently, `-fsSL` may be necessary. And `set -u` for *bash* might also help. 

---

Another trivial problem ~~for curiosity~~: why is *SauceCodePro ...* the default GUI font while it is not installed by default?

edrex:
Fixes #2532 

Also cleans up empty files so rerunning install.sh gets existing users in a good state

### PR Prelude

Fix for #2532 provided by @cebaa, also cleanup empty font files from that bug. The cleanup block can be removed if preferred.

### DIFF
--- a/docs/install.sh
+++ b/docs/install.sh
@@ -337,8 +337,13 @@ welcome () {
 
 # download_font {{{
 download_font () {
-  url="https://raw.githubusercontent.com/wsdjeg/DotFiles/master/local/share/fonts/$1"
+  url="https://raw.githubusercontent.com/wsdjeg/DotFiles/master/local/share/fonts/${1// /%20}"
   path="$HOME/.local/share/fonts/$1"
+  # Clean up after https://github.com/SpaceVim/SpaceVim/issues/2532
+  if [[ -f "$path" && ! -s "$path" ]]
+  then
+    rm "$path"
+  fi
   if [[ -f "$path" ]]
   then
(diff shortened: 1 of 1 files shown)


wsdjeg:
merged, thanks

PR #3069, opened 2019-09-19
Set search tools for Grepper

jyscao:
- set default search tools available to Grepper (before filtering)
  to be the same as the tools available to FlyGrep

### PR Prelude

I believe having the same set of search tools for both FlyGrep and Grepper is a more sensible default setting.

One suggestion I'd might add is that you should consider amending the list of default search tools defined by `g:spacevim_search_tools` in `.SpaceVim/autoload/SpaceVim.vim` to also include git-grep and sift, the only 2 tools supported by Grepper but not by FlyGrep (and have them fail silently when encountered by FlyGrep). This way, those who prefers to use git-grep or sift in Grepper can still do so.

### DIFF
--- a/config/plugins/vim-grepper.vim
+++ b/config/plugins/vim-grepper.vim
@@ -1,0 +1,9 @@
+" initialize g:grepper if it does not exist
+if !exists('g:grepper')
+  let g:grepper = {}
+endif
+
+" set value of g:grepper['tools'] to the same list of tools used by FlyGrep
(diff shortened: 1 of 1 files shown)


wsdjeg (approved):

wsdjeg:
How about support all of this in flygrep?

['git', 'ag', 'ack', 'ack-grep', 'grep', 'findstr', 'rg', 'pt', 'sift']

jyscao:
Excellent!

PR #3105, opened 2019-10-02
Preserve Git Hunks while using Github Layer

jyscao:
- check if github layer enabled
- if so, change Hunnks key from 'h' to 'H'
- otherwise, leave it as 'h' for compat

### PR Prelude

Currently, when the Github layer is enabled, the "+Hunks" functionality under the Git layer disappears. This PR fixes that.

### DIFF
--- a/autoload/SpaceVim/layers/git.vim
+++ b/autoload/SpaceVim/layers/git.vim
@@ -85,12 +85,14 @@ function! SpaceVim#layers#git#config() abort
   call SpaceVim#mapping#space#def('nnoremap', ['g', 'V'], 'GV!', 'git-log-of-current-file', 1)
   call SpaceVim#mapping#space#def('nnoremap', ['g', 'v'], 'GV', 'git-log-of-current-repo', 1)
 
-  " If github layer enabled, change Hunks key from 'h' to 'H' to avoid its functionality from being overwritten
-  let git_hunks_key = SpaceVim#layers#isLoaded('github') ? 'H' : 'h'
-  let g:_spacevim_mappings_space.g[git_hunks_key] = {'name' : '+Hunks'}
-  call SpaceVim#mapping#space#def('nmap', ['g', git_hunks_key, 'a'], '<Plug>(GitGutterStageHunk)', 'stage-current-hunk', 0)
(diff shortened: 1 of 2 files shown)


wsdjeg:
How about use `h` for both github layer and git hunks, I mean change key bindings maybe make user confuse.

jyscao:
I'd considered this potential problem too. But imo, I think it's still
better to separate the two as they serve different purposes.

And this is why I made sure to do the if check. So that changes to what all
users see/experience are minimized.

Note that users who currently have the GitHub layer enabled are probably
not even aware of the Hunks submenu via SpaceVim's shortcuts as it has been
overwritten. With the change, they'll still be able to access GitHub layer
bindings exactly as they did before with 'h', but will also be able to
discover the Hunks submenu as a new feature via [SPC]-g-H

Maybe it's worth adding a note in the docs for the benefit of new users
that if one enables the GitHub layer in addition to the git layer, then the
Hunks submenu under git gets remapped from g-h to g-H

jyscao:
On closer inspection, I think your suggestion is likely better overall. Because I just realized that whether the Hunks or GitHub submenu gets overwritten depends on the order a user enables their `[[layers]]` in their `init.toml` config file. Even though in almost all cases, we'd expect users to have the git layer enabled prior to the github layer. But ofc that cannot be guaranteed or assumed.

So rather than doing complicated order-checking to preserve their present keybindings, it will be simpler and easier to maintain to just combine the GitHub and Hunks keybinds into the same submenu.

See my updated commit.

wsdjeg:
ok, thanks. I will check it tonight.



在 Jethro Cao <notifications@github.com>，2019年10月2日 下午3:42写道：

On closer inspection, I think your suggestion is likely better overall. Because I just realized that whether the Hunks or GitHub submenu gets overwritten depends on the order a user enables their [[layers]] in their init.toml config file. Even though in almost all cases, we'd expect users to have the git layer enabled prior to the github layer. But ofc that cannot be guaranteed or assumed.

So rather than doing complicated order-checking to preserve their present keybindings, it will be simpler and easier to maintain to just combine the GitHub and Hunks keybinds into the same submenu.

See my updated commit.

wsdjeg:
merged, thanks.

PR #3154, opened 2019-10-15
Add npm support for building 2 plugins

jyscao:
- with only 'yarn install', build fails on systems without yarn
- log error to SpaceVim if neither npm or yarn are found

### PR Prelude

### DIFF
--- a/autoload/SpaceVim/layers/lang/markdown.vim
+++ b/autoload/SpaceVim/layers/lang/markdown.vim
@@ -25,13 +25,22 @@ function! SpaceVim#layers#lang#markdown#plugins() abort
   call add(plugins, ['joker1007/vim-markdown-quote-syntax',{ 'on_ft' : 'markdown'}])
   call add(plugins, ['mzlogin/vim-markdown-toc',{ 'on_ft' : 'markdown'}])
   call add(plugins, ['iamcco/mathjax-support-for-mkdp',{ 'on_ft' : 'markdown'}])
+  call add(plugins, ['lvht/tagbar-markdown',{'merged' : 0}])
+  " check node package managers to ensure building of 2 plugins below
+  if executable('npm')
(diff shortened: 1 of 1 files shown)


autoload/SpaceVim/layers/lang/markdown.vim
```diff
@@ -25,13 +25,22 @@ function! SpaceVim#layers#lang#markdown#plugins() abort
...
   call add(plugins, ['iamcco/mathjax-support-for-mkdp',{ 'on_ft' : 'markdown'}])
+  call add(plugins, ['lvht/tagbar-markdown',{'merged' : 0}])
+  " check node package managers to ensure building of 2 plugins below
+  if executable('npm')
+    let s:node_pkgm = 'npm'
+  elseif executable('yarn')
+    let s:node_pkgm = 'yarn'
+  else
+    let s:node_pkgm = ''
+    call SpaceVim#logger#error('npm or yarn is required to build iamcco/markdown-preview and neoclide/vim-node-rpc')
```
wsdjeg, line 36: no need to log it， the message will be seen when run spinstall.

jyscao: Only a generic failed messages gets seen though, like this: `x markdown-preview.nvim: Building failed.`

And this is what currently gets logged in the SpaceVim runtime log: 
`[ SpaceVim ] [17:28:03] [ Warn ]  [ plug manager ] Reinstalling Failed Plugins. Remaining Retries: 1`

When the building for that plugin failed, I had to look up the docs for the actual plugin first before realizing that it was because I didn't have `yarn`. A more to the point error message in the log would be more helpful imo.

What do you think?

wsdjeg:
it seems a bug, it should be  `x markdown-preview.nvim: Building failed, yarn is not find`

PR #3164, opened 2019-10-19
change lang#java mapping-gd, add a new mapping

alanding1989:
1. **lang#java** layer`s  **gd**  mapping is wrong,  I update it. 

2. add a new **mapping** based on vim-javacomplete2.
pls check the changed file for details.


wsdjeg:
go_to_def is empty function?

alanding1989:
> go_to_def is empty function?

I don`t know which function should use when with no lsp support.  
I haven`t find any function in vim-javacomplete2

alanding1989:
> > go_to_def is empty function?
> 
> I don`t know which function should use when with no lsp support. I haven`t find any function in vim-javacomplete2

PR #3173, opened 2019-10-20
Change rust.vim

alanding1989:
### PR Prelude
**Cargo** is Rust\`s package manager like maven to Java.
1. add **Cargo** support.
2. add a layer option of **racer_cmd**

pls check the changed file for details.


wsdjeg:
I just check all lsp key binding in other language layer, many are different, I will merge this PR, and unify all shortcuts。

PR #3192, opened 2019-10-24
Improve window swapping

jyscao:
- swap windows without needing to choose when there are only 2 windows

### PR Prelude

Currently `SPC w M` always prompts users to choose a window to swap with your active one using the command `ChooseWinSwap` from the vim-choosewin plugin, even when there are no windows to be swapped with or only one other window.

This change fixes that, so when there are only 1 or 2 windows, the equivalent of `<C-w> x` is pressed.

### DIFF
--- a/autoload/SpaceVim/mapping/space.vim
+++ b/autoload/SpaceVim/mapping/space.vim
@@ -198,7 +198,7 @@ function! SpaceVim#mapping#space#init() abort
         \ , 1)
   let s:lnum = expand('<slnum>') + s:funcbeginline
   call SpaceVim#mapping#space#def('nnoremap', ['w', 'M'], 
-        \ "execute {-> winnr('$')<=2 ? 'wincmd x' : 'ChooseWinSwap'}()",
+        \ "execute eval(\"winnr('$')<=2 ? 'wincmd x' : 'ChooseWinSwap'\")",
         \ ['swap window',
         \ [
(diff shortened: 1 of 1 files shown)


autoload/SpaceVim/mapping/space.vim
```diff
@@ -197,7 +197,8 @@ function! SpaceVim#mapping#space#init() abort
         \ ]
         \ , 1)
   let s:lnum = expand('<slnum>') + s:funcbeginline
-  call SpaceVim#mapping#space#def('nnoremap', ['w', 'M'], 'ChooseWinSwap',
+  call SpaceVim#mapping#space#def('nnoremap', ['w', 'M'], 
+        \ "execute {-> winnr('$')<=2 ? 'wincmd x' : 'ChooseWinSwap'}()",
         \ ['swap window',
```
wsdjeg, line 202: lamada is not supported in old vim.

jyscao: do you prefer if I changed it to have this mod support all versions, or make the mod only available to versions of vim/nvim that have lambda support?

wsdjeg: in old vim, we can use `exe string`.

jyscao: hey bud, I changed the mapped command to use `eval()` instead of the lambda now

PR #3193, opened 2019-10-24
Support search key with prefix

blaisegarant:
### PR Prelude

Right now, one cannot have a search pattern for a file with a prefix. But by default, python's pyunit default filename format is either `test_script.py` or `testscript.py` 

If the key is in the format `path/prefix_*.ext`, then `substitute(key, '*', '**/*', 'g')` returns `path/prefix_**/*.ext` and therefore `CMP.globpath` is empty although the key is already a valid format.

Adding both search with and without sub let you take both formats.

### DIFF
--- a/autoload/SpaceVim/plugins/a.vim
+++ b/autoload/SpaceVim/plugins/a.vim
@@ -37,9 +37,10 @@ endfunction
 
 function! s:paser(conf, root) abort
   for key in keys(a:conf)
-    let with_sub = s:CMP.globpath('.', substitute(key, '*', '**/*', 'g'))
-    let no_sub = s:CMP.globpath('.', key)
-    for file in with_sub + no_sub
+    let searchpath = key
+    if match(key, '/*')
+        let searchpath = substitute(key, '*', '**/*', 'g')
+    for file in s:CMP.globpath('.', searchpath)
       let file = s:FILE.unify_path(file, ':.')
       if has_key(a:conf, file)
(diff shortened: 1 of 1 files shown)


wsdjeg:
a checker before run global function will be better. run globalpath twice is slow.

blaisegarant:
Something in the likes of this?

autoload/SpaceVim/plugins/a.vim
```diff
@@ -37,7 +37,10 @@ endfunction
 
 function! s:paser(conf, root) abort
   for key in keys(a:conf)
-    for file in s:CMP.globpath('.', substitute(key, '*', '**/*', 'g'))
+    let searchpath = key
+    if match(key, '/*')
+        let searchpath = substitute(key, '*', '**/*', 'g')
```
wsdjeg, line 42: I think you miss an `endif`.

blaisegarant: Done, sry 'bout that

wsdjeg:
I still can not understand this PR. when the source code path is `src/foo.py` and the test script path is `test/test_foo.py`, how does you jump between them?

wsdjeg:
Ok, thanks, merged, and the config is:

```json
{
  "src/*.py": {"alternate": "test/test_{}.py"},
  "test/test_*.py": {"alternate": "src/{}.py"}
}
```

PR #3260, opened 2019-12-19
Add: Spacevim#custom#LangSPC and SpaceVim#custom#LangSPCGroupName

SleepyBag:
with these two interface, users are able to customize language specific commands (SPC+l) for any file types. For example:
```
  call SpaceVim#custom#LangSPCGroupName('python', ['d'], '+Debug')
  call SpaceVim#custom#LangSPC('python', 'nmap', ['d', 's'], 'echo "good"', 'let vim say good', 1)

```
defines a SPC+l+d group names '+Debug', and a key binding SPC+l+d+s for 'echo "good" ', exclusively for python

### DIFF
--- a/autoload/SpaceVim.vim
+++ b/autoload/SpaceVim.vim
@@ -725,13 +725,13 @@ let g:spacevim_filemanager             = 'vimfiler'
 " @section filetree_direction, options-filetree_direction
 " @parentsection options
 " Config the direction of file tree. Default is 'right'. you can also set to
-" 'left'. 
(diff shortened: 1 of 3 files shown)


wsdjeg:
thanks, I will test it. BTW doc for this two function is missing.

SleepyBag:
Thank you for your reply.
I will work on the document and append on this PR if it's going to be merged. ;-)

kifirkin:
Definitely must have feature :+1:

PR #3262, opened 2019-12-20
Add plugin to enable search via */# in visual mode

jyscao:
It's a useful addition that makes `*` and `#` work as one'd intuitively expect in visual mode.

the `incsearch` layer seemed like the most fitting place to add the plugin. Let me know if you think there's a better location; while convenient, the functionality itself doesn't seem critical enough for it to be added into core.


wsdjeg:
Merged, thanks.

PR #3271, opened 2019-12-30
feat: optional support of vim-devicons

blackxored:
### PR Prelude

The default icon API present in files supports very few icons. I see my React TSX files as TypeScript for example. [vim-devicons](https://www.github.com/ryanoasis/vim-devicons) supports a lot more. Unfortunately, SpaceVim makes this plugin useless due to its heavy customization (icons, airline formatter, etc). This PR addresses the icon part by optionally choosing to use devicons if the user has installed it.
By changing the SpaceVim's API we ensure that both startify and tabline work with this.

### DIFF
--- a/autoload/SpaceVim/api/file.vim
+++ b/autoload/SpaceVim/api/file.vim
@@ -141,6 +141,9 @@ let s:file_node_pattern_matches = {
 
 
 function! s:filetypeIcon(path) abort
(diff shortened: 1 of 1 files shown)


wsdjeg:
Instead of using vim-devicons, add icons to the api will be better.

PR #3299, opened 2020-01-23
Improve scheme layer

### LINKED ISSUE #3298: Support Chez Scheme dialect
**What is language?**

lang#scheme

**Which platform does the language support?**

- [x] windows
- [x] Mac os
- [x] Windows

**Any suggest vim plugins for this language?**

No

**Additional context**

[Chez Scheme](https://cisco.github.io/ChezScheme/) is a Scheme dialect by Cisco Inc.  Its executable named 'scheme' at Windows, but named 'chez' at HomeBrew.  Because SpaceVim doesn't support conditional layer option, it prevents  one ``init.toml`` to be used at two OS.

wsdjeg:
close #3298 

Whit this PR, users can config the scheme layer via:

```
[[layers]]
    name = 'lang#scheme'
    scheme_interpreter = 'C:\Program Files\Chez Scheme 9.5\bin\ta6nt\scheme.exe'
```

### Chez Scheme support

**REPL**

chez scheme repl

**Code runner**

Chez Scheme runner

### DIFF
--- a/autoload/SpaceVim/layers/lang/scheme.vim
+++ b/autoload/SpaceVim/layers/lang/scheme.vim
@@ -13,18 +13,19 @@ if exists('s:scheme_interpreter')
   " when call an autoload func, vim will try to load the script again
   finish
 else
-  let s:scheme_interpreter = 'scheme'
+  let s:scheme_interpreter = ''
+  let s:scheme_dialect = ''
 endif
 
 
 function! SpaceVim#layers#lang#scheme#config() abort
-  if s:scheme_interpreter ==# 'mit-scheme'
-    call SpaceVim#plugins#runner#reg_runner('scheme', 'echo | mit-scheme --quiet --load %s && echo')
+  if s:scheme_dialect ==# 'mit-scheme'
+    call SpaceVim#plugins#runner#reg_runner('scheme', 'echo | ' . shellescape(s:scheme_interpreter) . ' --quiet --load %s && echo')
     call SpaceVim#plugins#repl#reg('scheme', [s:scheme_interpreter, '-q'])
-  elseif s:scheme_interpreter ==# 'guile'
-    call SpaceVim#plugins#runner#reg_runner('scheme', 'echo | guile -q %s && echo')
+  elseif s:scheme_dialect ==# 'guile'
+    call SpaceVim#plugins#runner#reg_runner('scheme', 'echo | ' . shellescape(s:scheme_interpreter) . ' -q %s && echo')
     call SpaceVim#plugins#repl#reg('scheme', [s:scheme_interpreter, '-q'])
-  elseif s:scheme_interpreter =~? 'chez'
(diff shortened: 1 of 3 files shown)


thawk:
This solution doesn't seem to work with the HomeBrew version of Chez Scheme.
1. The executable file of Chez Scheme in HomeBrew is ``/usr/local/bin/chez``, I don't think it can match with ``Chez Scheme``.
2. The interpreter of other two dialect is basename, so it's better to config Chez Scheme with basename instead of full path.
3. Chez Scheme doesn't support ``--silent`` parameter.  It supports a ``--quiet``.  BTW, I check the manual of mit-scheme and guile, but I can't found the ``--silent`` parameter, maybe it's a typo?
4. Maybe the combine of original ``dialect`` option, and additional ``runner_option``/``repl_option`` can be a better solution?  Because the options required by the runner and by the repl are different (e.g. ``--scripts %s`` for runner, ``--quiet %s`` for repl).

wsdjeg:
@thawk 

1. change the expr to `=~? 'chez'` , then both `chez` and `Chez` will be matched.

thawk:
I don't know what's the executable name in all other OS, like different Linux distro. Maybe they use ``/usr/bin/scheme`` as in ArchLinux([PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=chez-scheme)). So it can't match anything about ``chez``.
I want to have a solution to share the config between different computers. I use the same configuration between Windows/MacOS/Linux/FreeNAS and the executable paths  are different between them.

wsdjeg:
>  I want to have a solution to share the config between different computers. 

support conditional layer option will be done in next release, before this is done, you need to use bootstrap function to config the layer.

wsdjeg:
> Maybe they use /usr/bin/scheme as in ArchLinux(PKGBUILD).
@thawk This is not officail package,  it is aur package.

thawk:
> > Maybe they use /usr/bin/scheme as in ArchLinux(PKGBUILD).
> > @thawk This is not officail package,  it is aur package.

[Ubuntu](http://manpages.ubuntu.com/manpages/bionic/man1/chezscheme9.5.1.html) uses ``scheme`` too.

thawk:
> 》 I want to have a solution to share the config between different computers.
> 
> support conditional layer option will be done in next release, before this is done, you need to use bootstrap function to config the layer.

It's good to know it.  Should I using my ``autocmd FileType scheme`` to override the runner and repl? Or do I miss something?

wsdjeg:
No, just using following snippet in bootstrap after function.

```
if  .....
       call SpaceVim#plugins#runner#reg_runner('scheme', '.....')
       call SpaceVim#plugins#repl#reg('scheme', '....')
else
       call SpaceVim#plugins#runner#reg_runner('scheme', '.....')
       call SpaceVim#plugins#repl#reg('scheme', '....')
endif

```

PR #3300, opened 2020-01-23
Changes plugins to forked version

wsdjeg:


crisidev:
Hello, thanks for the awesome job! What is the rationale behind this change? How would you ensure these security checks? What will they be? 

I am quite concerned about this change as now any update to plugins will have to pass through you..

wsdjeg:
Closed, as we are using bundle plugins now, keep all plugin in box, make it easier to install and debug.

PR #3322, opened 2020-02-01
Improve lang#r layer

### LINKED ISSUE #3287: Python and R IDEs repl does not work as expected on Termux.
https://github.com/SpaceVim/SpaceVim/issues/2754#issuecomment-573274002

SPC l r 
does not run the script within the REPL.

Another problem with lang#r:

Repl runs not in the same directory as file but its parent directory. This causes problem when a program expects input files in the same directory as R script.

```toml

#============================= == == == == == ============
#dark_powered.toml-- - dark powered configuration example for SpaceVim
#Copyright(c) 2016 - 2017 Wang Shidong &Contributors
#Author : Wang Shidong < wsdjeg at 163.com>
#URL : https: // spacevim.org
#License : GPLv3
#============================================================
#All SpaceVim option below[option] section
# This is a basic configuration example for SpaceVim

# All SpaceVim options are below [options] snippet
[options]
    # set spacevim theme. by default colorscheme layer is not loaded,
    # if you want to use more colorscheme, please load the colorscheme
    # layer, the value of this option is a string.
    colorscheme = "gruvbox"
    colorscheme_bg = "dark"
    # Disable guicolors in basic mode, many terminal do not support 24bit
    # true colors, the type of the value is boolean, true or false.
    enable_guicolors = true
    # Disable statusline separator, if you want to use other value, please
    # install nerd fonts
    statusline_separator = "nil"
    statusline_separator = "bar"
    buffer_index_type = 4
    # Display file type icon on the tabline, If you do not have nerd fonts
    # installed, please change the value to false
    enable_tabline_filetype_icon = true
    # Display current mode text on statusline, by default It is disabled,
    # only color will be changed when switch modes.
    # checkers options
    enable_statusline_mode = false
    enable_neomake = false
    enable_ale = true
    lint_on_the_fly = true

# Enable autocomplete layer
[[layers]]
    name = "autocomplete"
    auto-completion-return-key-behavior = "complete"
    auto-completion-tab-key-behavior = "cycle"

[[layers]]
    name = "shell"
    default_position = "top"
    default_height = 30

# This is an example for adding custom plugins lilydjwg/colorizer
[[custom_plugins]]
    name = "lilydjwg/colorizer"
    merged = false


[[layers]] 
    name = "lang#java"

[[layers]] 
    name = "lang#python"
    python_file_head = [
      '#!/usr/bin/env python',
      '# -*- coding: utf-8 -*-',
      '',
      ''
    ]
    enable_typeinfo = true

[[layers]] 
    name = "lang#javascript"

[[layers]] 
    name = "lang#perl"

[[layers]] 
    name = "lang#ruby"

[[layers]] 
    name = "lang#c"

[[layers]] 
    name = "lang#r"
    repl_command = "R --vanilla"

[[layers]] 
    name = "format"

[[layers]]
    name = "checkers"

[[layers]] 
    name = 'shell' 
    default_position = 'top' 
    default_height = 30
```

The above is my init.toml configuration under directory .SpaceVim.d.

This isn't exactly per the issue template. Do let me know if you need any additional info.

wsdjeg:
close #3287

### PR Prelude

### DIFF
--- a/autoload/SpaceVim/layers/lang/r.vim
+++ b/autoload/SpaceVim/layers/lang/r.vim
@@ -18,14 +18,15 @@ function! SpaceVim#layers#lang#r#set_variable(var) abort
 endfunction
 
 function! SpaceVim#layers#lang#r#config() abort
-  call SpaceVim#plugins#runner#reg_runner('r', 'R --vanilla <%s')
+  call add(g:spacevim_project_rooter_patterns, '.Rprofile')
+  call SpaceVim#plugins#runner#reg_runner('r', 'R <%s')
   call SpaceVim#mapping#space#regesit_lang_mappings('r', function('s:language_specified_mappings'))
   if !empty(s:r_repl_command)
-      call SpaceVim#plugins#repl#reg('r',s:r_repl_command)
+    call SpaceVim#plugins#repl#reg('r',s:r_repl_command)
   else
-      call SpaceVim#plugins#repl#reg('r', 'r')
+    call SpaceVim#plugins#repl#reg('r', 'r')
   endif
-  
+
 endfunction
 
 function! s:language_specified_mappings() abort
--- a/autoload/SpaceVim/plugins/projectmanager.vim
+++ b/autoload/SpaceVim/plugins/projectmanager.vim
@@ -168,7 +168,9 @@ function! s:find_root_directory() abort
       let dir = SpaceVim#util#findFileInParent(pattern, fd)
     endif
     let ftype = getftype(dir)
-    if ( ftype ==# 'dir' || ftype ==# 'file' ) && dir !=# expand('~/.SpaceVim.d/')
+    if ( ftype ==# 'dir' || ftype ==# 'file' ) 
+          \ && dir !=# expand('~/.SpaceVim.d/')
+          \ && dir !=# expand('~/.Rprofile')
       let dir = s:FILE.unify_path(fnamemodify(dir, ':p'))
       if ftype ==# 'dir'
         let dir = fnamemodify(dir, ':h:h')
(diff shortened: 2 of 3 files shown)


linusjf:
It's my understanding that the R environment should be run from the directory in which the .Rprofile file exists. This does not have to be the directory from which SpaceVim was executed. If the .Rprofile does not exist in this directory, SpaceVim can look at the parent directory to check if it exists there and  change the working directory to that if it does. If it doesn't, the R environment will use the .Rprofile from the user home directory or the R installation, in that order. What should the working directory be in the latter two cases? I'd assume it has to be the directory from which SpaceVim was executed.

wsdjeg:
I have add .Rprofile to the project root partten, and SpaceVim will find .Rprofile file in the parent directory of the openned file.

linusjf:
I'm not sure I understand you. Do you always need a .Rprofile file? What if it doesn't exist? Shouldn't the working directory then be the directory from which vim was executed? I'm not familiar with the vim scripting language.

wsdjeg:
I mean I have add .Rprofile into spacevim project root pattern, SpaceVim find prject based on the project root pattern, if no root path is found, the path will not be changed. the project_root_pattern is a list, ['.git/', 'package.json', '.Rprofile']

linusjf:
Are all these patterns used for the R layer? What if I don't want any of these patterns to be used? Specifically, in my case, no .Rprofile exists but .git does. But I don't want .git directory to be the working dir. Then what?

linusjf:
I tried the latest version of SpaceVim.
However, SPACE l r still doesn't function as expected.

This time, R is executed but without one of the required options: --save, --no-save, and --vanilla.

Space l s , on the other hand, executes R --no-save. 

Where are the parameters for SPACE l r picked up from? Ought these not to be the same as in the REPL specified in init.toml?

PR #3338, opened 2020-02-06
Fix(lsp): multiple language server registration bug

swr1bm86:
### PR Prelude

if we register multiple language servers in our `init.toml` like following below:

```
[[layers]]
    name = "lsp"
    filetypes = [
        "rust",
        "c",
        "cpp",
    ]
    [layers.override_cmd]
        rust = ["rustup", "run", "nightly", "rls"]
        c = ["clangd"]
```

`vim-lsp` will always store the last language server instance since every time we register a new language server, we always use `LSP` as the instance name which is also the [index](https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp.vim#L151-L163) used inside `vim-lsp`, we can check this by `:call lsp#print_server_status()`

image

and this simple fix is to change the name of each lsp instance from fixed `LSP` to `filetype-lsp`, e.g. `rust-lsp`, `cpp-lsp`, etc, and then everything just goes well as expected :)

image

### DIFF
--- a/autoload/SpaceVim/lsp.vim
+++ b/autoload/SpaceVim/lsp.vim
@@ -114,7 +114,7 @@ else
   " use vim-lsp
   function! SpaceVim#lsp#reg_server(ft, cmds) abort
     exe 'au User lsp_setup call lsp#register_server({'
-          \ . "'name': 'LSP',"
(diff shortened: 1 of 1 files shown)


wsdjeg:
awesome, merged.

PR #3340, opened 2020-02-07
Show buffer name

### LINKED ISSUE #3339: Tab manager should use buffer name as default tab name
When using the ``junkblocker/patchreview-vim`` plugin to do code review, it opens many tabs, one for a file.  Because tab count is more than 10, can't switch directly.  And because the tab name is empty, It's hard to swtich to the tab I want in the tab manager.

The tabline use buffer name if tab name is not set, but the tab manager doesn't has a default tab name.  Use the same same for tabline and tab manager is convenient for user.

wsdjeg:
close #3339

### DIFF
--- a/autoload/SpaceVim/layers/core/tabline.vim
+++ b/autoload/SpaceVim/layers/core/tabline.vim
@@ -98,6 +98,7 @@ function! SpaceVim#layers#core#tabline#get() abort
       let buflist = tabpagebuflist(i)
       let winnr = tabpagewinnr(i)
       let bufname = bufname(buflist[winnr - 1])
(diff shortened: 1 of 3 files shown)


thawk:
It seems that using ``fnamemodify(..., ":t")`` like in ``tabline.vim`` is better.

wsdjeg:
yeah, but how can we deal with plugin buffer?

thawk:
Does ``plugin buffer`` reference for buffers like ``__Tagbar__`` or ``Tab Manager``?   Could we distinct them from normal buffer?  At least, we can use what we see at the tabline, it may not perfect, but better than now.

PR #3348, opened 2020-02-14
Add file copy key binding

wsdjeg:


xylinkun:
The Key binding <kbd>SPC f Y</kbd> cannot work.
image

PR #3381, opened 2020-02-25
Add: lang/layer/go `GoFmt` shortcut

swr1bm86:
### PR Prelude

currently, we only got `GoImports` cmd to format imports without `GoFmt` cmd to format the current Go file.


wsdjeg:
Thanks for your PR, but in SpaceVim, we use `SPC b f` to format buffer, which is provided by format layer.

swr1bm86:
Ok, I see the details here https://github.com/sbdchd/neoformat/blob/master/autoload/neoformat/formatters/go.vim#L2. but from the go language layer, we will already got `goimports` to format imports with `SPC l m`, I think it will be super nice if we provide `gofmt` through language layer also, for the sake of completeness :)

wsdjeg:
sorry, we must keep all language layer has same key bindings, `SPC b f` is to format buffer for all languages, no need to add another language specific key binding.

PR #3400, opened 2020-03-10
Fix the git log commands for chosen git plugins

mprihoda:
### PR Prelude

The mappings to git log has been set to use the wsdjeg/git.vim, which does not have to be always installed. When using Gina, fugitive or Gita, the commands would fail with "ambiguous user command" message. I moved the mapping to the respective sections, updated the correct commands and fixed one type while at it.


wsdjeg:
merged, thanks.

PR #3411, opened 2020-03-17
Add: Respect $XDG_CACHE_HOME for the data/cache dir

jonfin:
### PR Prelude

On my companies PC the home directory stays on a network drive and some times the network throughput getting low. Therefore the access to the `$HOME/.cache` dir is slow and neovim is getting slower. I want to be able to change the path, so the path can point to a local directory.

The path `$HOME/.cache` is now configurable through the `$XDG_CACHE_HOME` environment variable or could also be set with the variable `g:spacevim_data_dir`. If the later is used, there are some plugins which will still write to `$XDG_CACHE_HOME`.

Even so it adds a new option/functionality, the default behavior is not changed, because the environment variable `$XDG_CACHE_HOME` point normally to `$HOME/.cache`. If `$XDG_CACHE_HOME` is not set, it uses as default `$HOME/.cache`.

### DIFF
--- a/autoload/SpaceVim.vim
+++ b/autoload/SpaceVim.vim
@@ -179,20 +179,41 @@ let g:spacevim_windows_leader          = 's'
 let g:spacevim_enable_insert_leader    = 1
 
 ""
+" @section data_dir, options-data_dir
+" @parentsection options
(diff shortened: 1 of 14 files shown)


wsdjeg:
merged, thanks

PR #3469, opened 2020-04-23
Solargraph does not use .bat extension

imanel:
### PR Prelude

Solargraph by default is providing executable without .bat extension, and correct version is even stated [in layer#lsp documentation](https://spacevim.org/layers/language-server-protocol/). However it looks like by accident it was added with incorrect extension, causing “unable to open solargraph.BAT” error. This PR removes this extension, and makes Ruby LSP work again.


wsdjeg:
It has `.BAT` extension in windows os.

imanel:
Most of Ruby programmers I know are working on Linux/macOS, but if changing this is not an option then would it be possible to at least update [layer#lsp configuration docs](https://spacevim.org/layers/language-server-protocol/#configuration) to show what is the command that is currently used (it's `solargraph.BAT` and not `solargraph` as in docs)? This should reduce amount of time needed to figure out why it's not working out of the box.

PR #3488, opened 2020-04-26
Doc: type in doc

previsualconsent:
simple typo in docs


wsdjeg:
I do not know if this is a type, because someone change it to "SourceCodePro" from "SauceCodePro".

O-XiXu:
this is about the default font for gui, like nvim-qt,
here in 'docs/documentation.md', it's correct, but not in ginit.vim, it should be 'Source' not 'Sauce'!
this give me a error when i run the gui :)

PR #3489, opened 2020-04-27
Fix projectmanager from finding ~/SpaceVim.d in home directory on windows

previsualconsent:
the project manager detects ~/SpaceVim.d in windows as a top level project. 

The current check for catching this is here:
https://github.com/SpaceVim/SpaceVim/blob/master/autoload/SpaceVim/plugins/projectmanager.vim#L172 

It compares the **strings** not the **paths** so on windows where paths use backslash, this comparison fails. 

I set the option **setslash** and also expand the path when comparing.

### DIFF
--- a/autoload/SpaceVim/default.vim
+++ b/autoload/SpaceVim/default.vim
@@ -126,6 +126,11 @@ function! SpaceVim#default#options() abort
   set shortmess+=s
   " Do not wrap lone lines
   set nowrap
+  
+  " set shellslash for filename comparisons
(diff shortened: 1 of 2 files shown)


wsdjeg:
we should not change the option setslash.

wsdjeg:
@previsualconsent Do you have a try with lasted version, I can not reproduce it now.

wsdjeg:
@previsualconsent This has been fixed in master branch, please update spacevim.

PR #3520, opened 2020-05-10
Fix coc.nvim installation.

tjni:
### PR Prelude

The installation scripts for coc.nvim have been removed.

### DIFF
--- a/autoload/SpaceVim/layers/autocomplete.vim
+++ b/autoload/SpaceVim/layers/autocomplete.vim
@@ -64,11 +64,7 @@ function! SpaceVim#layers#autocomplete#plugins() abort
           \ 'loadconf' : 1,
           \ }])
   elseif g:spacevim_autocomplete_method ==# 'coc'
-    if s:SYS.isWindows
-      call add(plugins, ['neoclide/coc.nvim',  {'merged': 0, 'build': './install.cmd'}])
-    else
-      call add(plugins, ['neoclide/coc.nvim',  {'merged': 0, 'build': './install.sh'}])
-    endif
+    call add(plugins, ['neoclide/coc.nvim',  {'merged': 0, 'rev': 'release'}])
   elseif g:spacevim_autocomplete_method ==# 'deoplete'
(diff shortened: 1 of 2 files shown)


wsdjeg:
so why can not use master branch?

tjni:
No strong reason. The release branch was what was recommended in the coc.nvim README, but pulling from master works as well. Please let me know which you prefer.

wsdjeg:
we should use `yarn install --frozen-lockfile` instead.

tranhoangcore:
#3540

tjni:
I tested using the `yarn` command, and I don't like that it doesn't work if `yarn` isn't installed. It also requires a build step each time coc.nvim is updated. For those reasons, I prefer keeping it on the release branch, which includes a pre-built `build` directory. What do you think?

This does require updating the plugin installation logic to switch branches properly, which I've changed in the latest commit.

Guzzii:
This is awesome. I was going to make the same PR.

@tjni But for some reason, it would not work by simply using the release like this `{'merged': 0, 'rev': 'release'}`. It would throw an error saying `javascript file not found`. The only way that works for me right now is to build from source with `{'merged':0, 'build': 'yarn install --frozen-lockfile'}`

tjni:
@Guzzii 

Would you be able to try with the most recent commit? It has a few additional lines updated that I expect to fix issues related to using `rev`. If it doesn't work, we can work on narrowing down where the problem is.

Guzzii:
> @Guzzii
> 
> Would you be able to try with the most recent commit? It has a few additional lines updated that I expect to fix issues related to using `rev`. If it doesn't work, we can work on narrowing down where the problem is.

Yep @tjni. I will try to give it shot today once I get some free time.

wsdjeg:
The release branch history is not same as the master branch, I just check the gitlog of coc, I think the most safe way to use this plugin is use build command. I will update the coc build command.

wsdjeg:
merged in 3564

PR #3534, opened 2020-05-16
Improve tools#mpv layer

wsdjeg:


autoload/SpaceVim/layers/tools/mpv.vim
```diff
@@ -6,16 +6,53 @@
 " License: GPLv3
 "=============================================================================
 
+if exists('s:musics_directory')
+  finish
+else
+  let s:musics_directory = '~/Musics'
```
goetzc, line 12: In plural? I think `~/Music` is more of a standard name.

wsdjeg:
The position can be changed in your toml file. the default can not be changed as it has been merged into master branch.
--------------
wsdjeg
***@***.*** commented on this pull request.

autoload/SpaceVim/layers/tools/mpv.vim
goetzc: `~/Music` is standard on macOS and also on FreeDesktop's XDG.
https://cgit.freedesktop.org/xdg/xdg-user-dirs/tree/user-dirs.defaults#n10

wsdjeg:
Ok，I will change it. thanks for your tips.

PR #3541, opened 2020-05-18
fix Leaderf neoyank.

mrbeardad:
And provide `nnoremap <C-L>` to refresh statusline.
And fix dirctory name of Ultisnips.

### PR Prelude

* `neoyank#update()` need to be called  before execute `Leaderf neoyank`
* `<C-L>` will refresh statusline. Some shortcut in neovim will trigger `WinLeave` but not `WinEnter`.
    In fact, you just want to type `<C-W>o` to close other windows. Then, statusline has been changed becase `WinLeave`
* the right dir variable is `UltiSnipsSnippetDirectories` , not `UltiSnipsSnippetsDir`

### DIFF
--- a/autoload/SpaceVim/layers/core/statusline.vim
+++ b/autoload/SpaceVim/layers/core/statusline.vim
@@ -568,6 +568,7 @@ function! SpaceVim#layers#core#statusline#init() abort
   augroup END
 endfunction
 
+nnoremap <silent><c-l> <c-l>:let &l:statusline = SpaceVim#layers#core#statusline#get(1)<cr>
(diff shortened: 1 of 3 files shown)


wsdjeg:
Thanks for your PR, but I can not understand why we need to map `<Ctrl-l>` the statusline should be updated based on autocmd.

mrbeardad:
> Thanks for your PR, but I can not understand why we need to map `<Ctrl-l>` the statusline should be updated based on autocmd.

I'm sorry that I forgeted to close this not robust PR.

The map `<c-l>` is to solve the problem that statusline doesn't update after some operator like `<c-w>o` and `<c-c>`. Maybe it is not a problem for the most.

PR #3639, opened 2020-07-17
Fix lua api

wsdjeg:
reproduce step:

1. switch to  branch `wsdjeg/lua_buf_api`
2. run `call SpaceVim#api#import('notification').notification('sssssssssssssssssss', 'WarningMsg')`

instead of show notification, the first line of current buffer will be changed, this only happened in gvim with `+lua` feature. I am not sure if this is a bug of vim's lua mode.

### DIFF
--- a/autoload/SpaceVim/api/notification.vim
+++ b/autoload/SpaceVim/api/notification.vim
@@ -116,8 +116,6 @@ function! s:self.redraw_windows() abort
     return
   endif
   let self.notification_width = max(map(deepcopy(self.message), 'strwidth(v:val)'))
-  call self.__buffer.buf_set_lines(self.border.bufnr, 0 , -1, 0, self.draw_border(self.title, self.notification_width, len(self.message)))
-  call self.__buffer.buf_set_lines(self.bufnr, 0 , -1, 0, self.message)
   let self.begin_row = 2
   for hashkey in keys(s:notifications)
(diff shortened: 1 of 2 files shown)


lilydjwg:
This is a bug.

To repro, just source:

```vim
let bufnr = bufadd('')
lua vim.buffer(vim.eval("bufnr"))[1] = "Ahh"
```

Here in if_lua.c, Vim temporarily sets `curbuf` to the buffer being updated:

https://github.com/vim/vim/blob/f0b9f43c3169037b7f2cd70e4939ba1ef26ada1c/src/if_lua.c#L1470-L1476

And in memline.c, `open_buffer` is called because this is a new buffer:

https://github.com/vim/vim/blob/f0b9f43c3169037b7f2cd70e4939ba1ef26ada1c/src/memline.c#L3333

Which save and restore stuff:

https://github.com/vim/vim/blob/f0b9f43c3169037b7f2cd70e4939ba1ef26ada1c/src/buffer.c#L365-L377

But when restoring, `curbuf` is set to the one inside `curwin`, not the previous one:

https://github.com/vim/vim/blob/f0b9f43c3169037b7f2cd70e4939ba1ef26ada1c/src/autocmd.c#L1591

Back into `ml_replace_len`, the `curbuf` points to the one in the current window.

PR #3759, opened 2020-09-03
fix g:spacevim_project_rooter_outermost

### LINKED ISSUE #3758: g:spacevim_project_rooter_outermost has inverse effect
## Expected behavior, english is required

When ``g:spacevim_project_rooter_outermost``  is true, the innermost directory is used instead of the outermost one.

## The reproduce ways from Vim starting (Required!)

* Create a directory structure like this:

 ```
  outer/
    .git/
    inner/
      .svn/
      test.txt
  ```

* Execute ``vim outer/inner/test.txt`` with ``g:spacevim_project_rooter_outermost``==true 
* ``:pwd``

It seems that the condition inside ``s:compare`` in ``autoload/SpaceVim/plugins/projectmanager.vim`` is wrong.

```vim
function! s:compare(d1, d2) abort
  if g:spacevim_project_rooter_outermost
    return len(split(a:d2, '/')) - len(split(a:d1, '/'))
  else
    return len(split(a:d1, '/')) - len(split(a:d2, '/'))
  endif
endfunction
```

If ``d1="outer"`` and ``d2="outer/inner"``, when ``outermost=true``, -1 should be returned, but 1 is returned instead.

thawk:
### PR Prelude

fix #3758

### DIFF
--- a/autoload/SpaceVim/plugins/projectmanager.vim
+++ b/autoload/SpaceVim/plugins/projectmanager.vim
@@ -222,7 +222,7 @@ function! s:sort_dirs(dirs) abort
 endfunction
 
 function! s:compare(d1, d2) abort
-  if g:spacevim_project_rooter_outermost
+  if !g:spacevim_project_rooter_outermost
     return len(split(a:d2, '/')) - len(split(a:d1, '/'))
   else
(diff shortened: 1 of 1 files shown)


wsdjeg:
@thawk Thanks for your hotfix.

PR #3762, opened 2020-09-04
:bug: cache init.toml based on filename if linked

ghedamat:
### PR Prelude

This PR attempts to fix the following edgecase

If init.toml is a symlink to another file and that file timestamp is not
updated when the file is changed then the caching system breaks.

This happens in particular for users that rely on [NixOS](https://nixos.org/) or
[home-manager](https://github.com/rycee/home-manager) to
manage their vim configuration. This is because in Nix all files have
the same timestamp of the UNIX EPOCH. When generating a new
version of the file Nix will checksum the name (hence changing the name
the links point to) but the timestamp will ALWAYS be the same.

The approach I took is to cache using the filename behind the link (if a link exists). This way Nix users should be able to manage their `init.toml` using home-manager and nobody else should be affected.

Please let me know if this is an acceptable change and/or if there is a better approach I should take.

Thanks a lot!

### DIFF
--- a/autoload/SpaceVim/custom.vim
+++ b/autoload/SpaceVim/custom.vim
@@ -209,9 +209,9 @@ function! s:load_glob_conf() abort
   if filereadable(global_dir . 'init.toml')
     let g:_spacevim_global_config_path = global_dir . 'init.toml'
     let local_conf = global_dir . 'init.toml'
-    let local_conf_cache = s:FILE.unify_path(expand(g:spacevim_data_dir.'/SpaceVim/conf/init.json'))
+    let local_conf_cache = s:FILE.unify_path(expand(g:spacevim_data_dir.'/SpaceVim/conf/' . fnamemodify(resolve(local_conf), ':t:r') . '.json'))
     let &rtp = global_dir . ',' . &rtp
-    if getftime(local_conf) < getftime(local_conf_cache)
+    if getftime(resolve(local_conf)) < getftime(resolve(local_conf_cache))
       let conf = s:JSON.json_decode(join(readfile(local_conf_cache, ''), ''))
       call SpaceVim#custom#apply(conf, 'glob')
     else


autoload/SpaceVim/custom.vim
```diff
@@ -209,9 +209,9 @@ function! s:load_glob_conf() abort
   if filereadable(global_dir . 'init.toml')
     let g:_spacevim_global_config_path = global_dir . 'init.toml'
     let local_conf = global_dir . 'init.toml'
-    let local_conf_cache = s:FILE.unify_path(expand(g:spacevim_data_dir.'/SpaceVim/conf/init.json'))
+    let local_conf_cache = s:FILE.unify_path(expand(g:spacevim_data_dir.'/SpaceVim/conf/' . fnamemodify(resolve(local_conf), ':t:r') . '.json'))
```
wsdjeg, line 212: Thanks for this PR, I do not understand, why the file name should be changed to `fnamemodify(resolve(local_conf), ':t:r')  . '.json'`

ghedamat: Yeah it's confusing..
Let me try to explain 


If you use Nix (https://nixos.org) to manage your dot files like init.toml
The timestamp of the file is always the same and it's the Unix epoch time (1970)

Because of this the current caching strategy does not work 

In Nix instead every new version of the file changes name because it contains an hash of the file contents in the name 

So between different versions the symlink in
.Spacevim.d/init.toml points to different files 


If we change caching this way it will keep working for existing users with no change but for people like me that use Nix the cache will be busted because the filename has changed.


Does this clarify? I can try to provide a more detailed example if you want 

And yes. I agree this is a limited use case but my argument is that it won't make it worse for anyone else...

Thanks for reading!

wsdjeg:
merged, Thanks

PR #3782, opened 2020-09-07
Improve issue template

raw-bin:
### PR Prelude

- This change improves the grammar in the text within the ISSUE_TEMPLATE.
- In addition it encourages the submitter to provide references to relevant SpaceVim documentation in order to ensure that it has been read and the intent understood before filing a new issue.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,25 +1,31 @@
-<!-- Reporting a bug without following this issue template structure will result in the bug being automatically closed -->
+<!-- Following this template is mandatory to open an issue. Please use English -->
 
-*****[ Please use English for reporting the issue ]**
+## Describe the behavior that you expect from SpaceVim
 
-## Describe the expected behavior
+**[This is a mandatory section]**
 
-**[ This is a mandatory section. If possible, in addition to the description, please also provide a link to the relevant SpaceVim documentation that describes the expected behaviour/feature ]**
+If possible, in addition to the description, please also provide a link to the relevant SpaceVim documentation that describes the expected behaviour/feature.
 
 ## Describe the problem
 
-**[ This is a mandatory section. Please describe the observed problem in simple and clear terms ]**
+**[This is a mandatory section]**
+
+Please describe the observed problem in simple and clear terms.
 
 ## Describe steps to reproduce the problem
 
-**[ This is a mandatory section. Please list all the steps from the point the Vim session was started until the problem was observed ]**
+**[This is a mandatory section]**
 
-## Provide SpaceVim debug info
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,15 +1,31 @@
...
+
+**[This is a mandatory section]**
+
+Please list all the steps from the point the Vim session was started until the problem was observed, in simple and clear terms.
+
+## Provide debug information from SpaceVim
+
+**[This is a mandatory section]**
+
+From within SpaceVim, please press `SPC h I` (ie: press and release`space`, followed by pressing and releasing the `h` key, followed by pressing and releasing the `I` key (capital I)). This will make SpaceVim print out debug information and also copy it to the clipboard. Please paste this debug information below in entirety.
```
wsdjeg, line 25: it should be "press ~~and release~~ `space`, "

raw-bin: On my system, I have to 'tap' `space` (so press and release, rather than press and hold), then tap `h` and finally tap `I`. I assume that is the expected way to use these keys ?

Basically, I wanted to prevent `SPC h I` being interpreted as:

- Press space and keep it pressed, then
- press h and keep it pressed, then
- press I

Does that make sense or have I misunderstood something ?

Rom1deTroyes: Have you tried this sequence (press together <kbd>Space</kbd>+<kbd>h</kbd>+<kbd>Shift</kbd>+<kbd>I</kbd>, than release all keys) ?

On my GNU/Linux laptop, it opens `issue_report.md` in a new tab (so it basically does the same thing... an emacs way :see_no_evil:)

The question I see here is does `press` needs to be disambiguate between `press down and release the key` (witch are two different actions : for me it's a tap) and `press any key` (obviously you don't keep the key pressed for ever) ?

raw-bin: @Rom1deTroyes: Thanks fore the response.  I see that too!

FWIW: I tried spacemacs recently and even it uses the 'tap' dynamic. So whenever they say `SPC a b` they mean taps.

I agree with you - the question here is how best to disambiguate 'press'. Maybe I should just change the text to:

```From within SpaceVim, please tap the following in order: `SPC h I` (ie: tap `space`, then tap `h` and then tap `I` (uppercase `i`). This....```

Do you think that's clearer ?

wsdjeg: yeah, LGTM

raw-bin: Thanks @wsdjeg : I'll submit a new PR.

wsdjeg: Ok, but why not just push to this pr?

PR #3803, opened 2020-09-12
Update defx.nvim

wsdjeg:


wsdjeg:
@Shougo, I have update defx to latest version, but the filter feature only works for file in root dir, and do not scan sub directory. I think you can check out how chadtree implement the feature.

https://github.com/ms-jpq/chadtree

Shougo:
@wsdjeg I ignore your report.  Because you don't create the new issue with minimal config.
I don't work for non reproduce-able issues.

Shougo:
> I think you can check out how chadtree implement the feature.

The suggestion is no means.  Because I need to reproduce the problem.
We have no ESP skills.

Shougo:
Please compare your issue with the issue.
https://github.com/Shougo/defx.nvim/issues/257

Which is reproduce-able for me?

PR #3837, opened 2020-09-25
Add: new configuration options for go layer to take advantage of gopls features

vadimi:
### PR Prelude

For `gopls` users `vim-go` plugin offers some options to use it for formatting and imports processing:
https://github.com/fatih/vim-go/blob/master/doc/vim-go.txt#L1352

This PR adds new configuration options to `go` layer and preserves the default behaviour. New configuration options are:

- `go_fmt_command`
- `go_imports_autosave`
- `go_imports_mode`

Example config that I use:

```
[[layers]]
name = "lang#go"
go_fmt_command = 'gopls'
go_imports_mode = 'gopls'
go_imports_autosave = 1
```

### DIFF
--- a/autoload/SpaceVim/layers/lang/go.vim
+++ b/autoload/SpaceVim/layers/lang/go.vim
@@ -40,6 +40,10 @@
 " <
 
 
+let s:go_fmt_command = 'gofmt'
+let s:go_imports_autosave = 0
+let s:go_imports_mode = 'goimports'
+
 function! SpaceVim#layers#lang#go#plugins() abort
   let plugins = [['fatih/vim-go', { 'on_ft' : 'go', 'loadconf_before' : 1}]]
   if has('nvim') && g:spacevim_autocomplete_method ==# 'deoplete'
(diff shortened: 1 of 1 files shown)


wsdjeg:
the code format is provided by format layer. and for some language layer, we have `format_on_save` option, I think maybe we can use this option for auto import too.

vadimi:
my understanding is format layer is using neoformat that has no knowledge of gopls: https://github.com/sbdchd/neoformat/blob/master/autoload/neoformat/formatters/go.vim. Also gopls is the solution for many code tasks in go that does not require installing any other tools, it also works faster that goimports for example for autoimporting modules.

vadimi:
@wsdjeg any update on this? I'd love to use more gopls features that vim-go plugin exposes. If it's not the direction you want to do, can you please let me know how I can add additional vim-go configuration to my config?

wsdjeg:
Hi, before this PR is merged, you can use bootstrp function to write vim script. the vim-go configuration can be set in bootstrap function.

What should be done with this PR?

1. the language layer option is only for this language layer, so maybe we can remove `go_` prefix, and make the option name shorter.

2. instead of use `fmt_command = 'gopls'`, maybe `enabled_formatter` is better, the value can be a list. it should be same as `neoformat_go_enabled_formatter`.

3. `import_mode`, this option should be merged into formatter.

4. instead of using `imports_autosave`, maybe, we can use `format_on_save`. and we can also add `lint_on_save` for `lang#go` layer.

after all of these have been done, the `lang#go` layer can be loaded like this:

```toml
[[layers]]
  name = 'lang#go'
  enabled_formatter = ['gopls']
  enabled_maker = ['golint']
  lint_on_save = true
  format_on_save = true
```

BTW, the layer page doc also should be updated.

vuvgaG:
Whats up with this?

vadimi:
unfortunately I didn't have enough time to work on this, but I was able to use bootstrap functions to achieve the desired behavior, closing this PR

PR #3860, opened 2020-10-02
Fix project root detection to work with empty buffer

agriffis:
### PR Prelude

So that you can run "vim" in a project and use fzf to find a file.


wsdjeg:
@agriffis The default logic of spacevim project detection do not skip empty buffer in old version, I do not remember why I add `skip empty buffer` feature. When start vim via `vim` in a directory, vim will call root detection function. but no file is opened, and current buffer is empty, now it can not find the root. so this is a bug.

agriffis:
@wsdjeg Hmm, interesting. For me, when I open vim with no file, I want it to get the project root based on the working directory. Are you suggesting there's a different approach to consider?

wsdjeg:
merged, I have test it, and we should not skip empty buffer.

wsdjeg:
Thanks!

PR #3864, opened 2020-10-04
Add initial support for vim-codefmt alternative to neoformat

agriffis:
### PR Prelude

vim-codefmt has capabilities beyond neoformat; it can format ranges and individual lines. This will need additional keybindings but this pull request adds the initial support.

Formatting ranges and lines is useful when contributing to a large project, where you want to format your new or modified code, but don't want to make extensive changes to the overall file.

### Keybindings discussion

This pull request doesn't add any keybindings, but we should talk about what makes sense.

For keybindings, SpaceVim currently has `SPC bf`. This keybinding implies that formatting will always be full buffer, and doesn't extend well to formatting individual lines, etc.

Spacemacs uses mode-specific bindings:

- `SPC m==` format line
- `SPC m=b` format buffer
- `SPC m=f` format function

Personally I think we could shorten these in SpaceVim to:

- `SPC ==` format current line (like normal == but with SPC leader)
- `SPC =b` format current buffer (replace `SPC bf`)
- `SPC =` format visual range


wsdjeg:
Sorry, we can not replace `SPC b f`, becase this key binding is top used key binding in my workflow, and `=` is too far away on key board.

`SPC b f` should support visual mode, That means, you can select lines, and press `SPC b f`, it will only format selections.

Then, we need only one key binding for format code.

agriffis:
@wsdjeg I started to rebase this on your changes, but I'm a little uncertain about it. In neoformat there's a single command, but in codefmt there are two commands: `:FormatCode` (entire buffer) and `:FormatLines` (current line or visual range). Do you have a suggestion?

wsdjeg:
@agriffis sorry for late reply, I do not know how to wrap these two commands.

PR #3869, opened 2020-10-05
Use bundle fzf plugin

wsdjeg:


ypid:
I am new to SpaceVim but to me this looks like those changes could go upstream instead of forking. Am I missing something?

PR #3888, opened 2020-10-11
Fix job.send can handle multi-line string

### LINKED ISSUE #3887: job.send() shouldn't convert string to list
Using ``repl#send('raw')`` to send a multi-line string, it should send out a multi-line string.  But in the result, newlines are replaced by ``NUL``(``0x00``).

* ``:help chansend()`` says, **"any newlines in an item (of list) will be sent as NUL"**:

  > {data} may be a string, string convertible, or a list.  If
  > {data} is a list, the items will be joined by newlines; any
  > newlines in an item will be sent as NUL. 

* ``job.send()`` converts argument of type ``string`` into ``list`` for neovim:
  ```vim
  if self.nvim_job
    if has_key(self.jobs, a:id)
      if type(a:data) == type('')
        call jobsend(a:id, [a:data, ''])
      else
        call jobsend(a:id, a:data)
      endif
  ```

* Any raw string contains ``\n`` will be convert to a list by job api, thus all newlines are replaced by NULs by neovim

``job.send()`` should simply append a ``\n`` to string argument, it should works as expected.

thawk:
### PR Prelude

close #3887

Because ``job.send()`` will append a ``\n`` for neovim and vim, so it's unnecessary to append one in ``repl#send()``.

If we don't want to limit the argument type of ``repl#send('raw')``, the type check can be omitted too.

### DIFF
--- a/autoload/SpaceVim/api/job.vim
+++ b/autoload/SpaceVim/api/job.vim
@@ -275,7 +275,7 @@ function! s:self.send(id, data) abort dict
   if self.nvim_job
     if has_key(self.jobs, a:id)
       if type(a:data) == type('')
-        call jobsend(a:id, [a:data, ''])
+        call jobsend(a:id, a:data . "\n")
       else
         call jobsend(a:id, a:data)
       endif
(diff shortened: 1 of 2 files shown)


wsdjeg:
yeah maybe the `job.send` API should be improved.

PR #3900, opened 2020-10-15
todo manager: implement custom keywords

### LINKED ISSUE #3466: vim-todo not working without spacevim
## Expected behavior, english is required

https://github.com/wsdjeg/vim-todo  
it says it works without spacevim and automatically detaches. it doesn't  
because this repo has no issues i think its also related to spacevim (since its programmed for that) so i will try my best here since i think this is a simple powerful plugin

## The reproduce ways from Vim starting (Required!)

add the plugin with your plugin manager without spacevim  
`:OpenTodo` in the command line

## Debug info

i am on linux with neovim

## Screenshots

2020-04-21-210901_1254x244_scrot

romgrk:
Fixes https://github.com/SpaceVim/SpaceVim/issues/3466

To supply custom search keywords, the user must set two options:
 - `spacevim_todo_labels`: the labels as will be searched by the external tool (grep, ripgrep, ...)
 - `spacevim_todo_labels_pattern`: the *vim* pattern to match the labels inside a string

The second option is required in case the first one contains special regex stuff that isn't handled by vim's patterns. The example below shows how to configure it to match the keywords "TODO" & friends as they usually appear in source code.

Screenshot from 2020-10-14 21-26-23

Notes:
 - Untested because I don't use SpaceVim but tested [this one](https://github.com/wsdjeg/vim-todo/pull/3).
 - IMHO the default behavior should be to include "TODO" etc. "@todo" can be included as well even if its usage is limited. Let me know what you think.

### DIFF
--- a/autoload/SpaceVim/plugins/todo.vim
+++ b/autoload/SpaceVim/plugins/todo.vim
@@ -67,18 +67,7 @@ function! s:update_todo_content() abort
   let argv = [s:grep_default_exe] + 
         \ s:grep_default_opt +
         \ s:grep_default_expr_opt
-  " @fixme expr for defferent tools
-  " when using rg, [join(s:labels, '|')]
-  " when using grep, [join(s:labels, '\|')]
-  if s:grep_default_exe ==# 'rg'
-    let argv += [join(s:labels, '|')]
-  elseif s:grep_default_exe ==# 'grep'
-    let argv += [join(s:labels, '\|')]
-  elseif s:grep_default_exe ==# 'findstr'
-    let argv += [join(s:labels, ' ')]
-  else
-    let argv += [join(s:labels, '|')]
-  endif
+  let argv += [s:join_labels()]
   if s:SYS.isWindows && (s:grep_default_exe ==# 'rg' || s:grep_default_exe ==# 'ag' || s:grep_default_exe ==# 'pt' )
     let argv += ['.']
   elseif s:SYS.isWindows && s:grep_default_exe ==# 'findstr'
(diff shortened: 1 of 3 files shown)


breuerfelix:
image
it is working like charm! well even though in the list the todo is still in lowercase even thought it is uppercase in the files. but i think that really doesn't matter :D

wsdjeg:
I think we should not add two options for one feature. for people who want to use `TODO` as label, only need to use:

```toml
[options]
  todo_labels = ['TODO']
```

we should add `\b` arroud the label automatiblly installed of adding them in option.

spacevim has a `vim#regex` api. we do not need to add new option for it;


the following test is past.

```
Execute ( SpaceVim api: vim#key ):
  let regex = SpaceVim#api#import('vim#regex')
  AssertEqual regex.parser('(?i)abc', 0), '\v\cabc'
  AssertEqual regex.parser('(?-i)abc', 0), '\v\Cabc'
  AssertEqual regex.parser('(?:exp)', 0), '\v%(exp)'
  AssertEqual regex.parser('\a\f\v', 0), '\v%x07%x0C%x0B'
  AssertEqual regex.parser('[:word:]', 0), '\v[0-9A-Za-z_]'
  AssertEqual regex.parser('\bspace\b', 0), '\v\<space\>'
  AssertEqual regex.parser('if%@&ab', 0), '\vif\%\@\&ab'
  AssertEqual regex.parser('(?P<name>exp)', 0), '\v(exp)'
```

breuerfelix:
Well to be honest i would try to maintain this plugin a little bit decoupled from spacevim. there is no such plugin anywhere for vim currently. also it says that it does work without spacevim (which is true after this PR) so why not decouple it ? :)

(neo)vim users do not have these options to change the TODO keyword :)

romgrk:
The regex.parser is not working:

```vim
echo SpaceVim#api#import('vim#regex').parser('\btest\b', 0)
" => '\vtest'
```

romgrk:
I've made it work without that. It now takes a single option, `spacevim_todo_labels`.

I would like to express my frustration, the workflow to contribute to this plugin is way harder than it should be. The plugin should stand by itself and exposes whatever API it needs to make it play nice with SpaceVim. This setup makes it very tedious to contribute to the plugin from outside SpaceVim.

wsdjeg:
@romgrk Thanks for contribute to this plugin, as vim-todo repo is just a mirror repo detached by script in `.ci` directory. I do not want to maintain duplecate code in github.

breuerfelix:
@wsdjeg what about using vim-todo as a submodule ? would make maintaining it easier as a stand alone plugin

wsdjeg:
@barcharcraz I do not family with submodule, and I do not like submodule, It always break my main repo status.

wsdjeg:
another hotfix patch will be push to master soon, the syntax highlight is wrong and backword compatible is broken.

PR #3956, opened 2020-11-12
Add: introduce splitjoin

gbprod:
### PR Prelude

Introduce [splitjoin](https://github.com/AndrewRadev/splitjoin.vim) plugin in edit layer.


gbprod:
I'm not shure about the key bindings, I would like to add it to the "SPC menu", something like `SPC x j i` or `SPC J i` but I can't find a available and mnemonic one...

Also, I think it could be relevant to replace jplus plugin by the more complete [Join](https://github.com/sk1418/join), what do you think ?

wsdjeg:
yeah, I will check these two plugins, and we should not use `ctrl-j` key binding, becaseu this key binding is used in tmux layer, and some people also use it to jump to windows below current one. and maybe we can use `SPC j`, and the description of `SPC j` is `+Jump/Join/Split` now.

gbprod:
There's also the awesome splitjoin plugin https://github.com/AndrewRadev/splitjoin.vim

I will try to improve this PR !

gbprod:
I've updated to introduce splitjoin plugin as a layer.

I would like to add the "Join with delimiter" binding but maybe in another PR

wsdjeg:
`jplus` has been included in edit layer, how about add this plugin into edit layer, layer should be a group of plugins which provide relative functions,

autoload/SpaceVim/layers/edit.vim
```diff
@@ -32,6 +32,7 @@ function! SpaceVim#layers#edit#plugins() abort
         \ [g:_spacevim_root_dir . 'bundle/vim-jplus', { 'on_map' : '<Plug>(jplus' }],
         \ [g:_spacevim_root_dir . 'bundle/tabular',           { 'merged' : 0}],
         \ [g:_spacevim_root_dir . 'bundle/vim-better-whitespace',  { 'on_cmd' : ['StripWhitespace', 'ToggleWhitespace', 'DisableWhitespace', 'EnableWhitespace']}],
+        \ ['andrewradev/splitjoin.vim',{ 'merged' : 0, 'loadconf' : 1}],
```
gbprod, line 35: Maybe I should add this plugin to "bundle" ?

wsdjeg: Yeah, because this layer is enabled by default, all plugins in this layer should be included in bundle dir.

PR #4127, opened 2021-03-11
Add task problem matcher

### LINKED ISSUE #4126: Jump to error line in tasks manager
tasks manager can be used instead of ``:make``  to compile project, but when there're some compile error, we can't jump to error line like in quickfix window.

In VSCode, tasks has a property named [``problemMatcher``](https://code.visualstudio.com/docs/editor/tasks#_defining-a-problem-matcher), which can be used to match the problem to jump to error line.  I suggest tasks manager in SpaceVim support the jumping too.

**Describe alternatives you've considered**
 1. load the final output into quickfix window, it automatically support error jumping.  But it can only work after the task is completely finished, which may need minutes for large C++ project.
 2. support ``errorformat`` natively in tasks manager, so the jumping is possible while the task is running, no need to wait it finish.

**Additional context**

wsdjeg:
close #4126

### DIFF
--- a/autoload/SpaceVim/plugins/runner.vim
+++ b/autoload/SpaceVim/plugins/runner.vim
@@ -409,7 +409,7 @@ function! s:match_problems(output, matcher) abort
     let olderrformat = &errorformat
     if has_key(a:matcher, 'errorformat')
       let &errorformat = a:matcher.errorformat
-      let cmd = 'noautocmd caddexpr a:output'
+      let cmd = 'noautocmd cexpr a:output'
       exe cmd
       copen
     endif
--- a/autoload/SpaceVim/plugins/runner.vim
+++ b/autoload/SpaceVim/plugins/runner.vim
@@ -411,6 +411,7 @@ function! s:match_problems(output, matcher) abort
       let &errorformat = a:matcher.errorformat
       let cmd = 'noautocmd cexpr a:output'
       exe cmd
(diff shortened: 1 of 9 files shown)


autoload/SpaceVim/plugins/runner.vim
```diff
@@ -392,16 +395,63 @@ function! SpaceVim#plugins#runner#run_task(task) abort
...
 
+function! s:match_problems(output, matcher) abort
+  if has_key(a:matcher, 'pattern')
+    let items = []
+    for line in a:output
+      let rst = matchlist(line, a:matcher.pattern.regexp)
+      let file = rst[a:matcher.pattern.file]
+      let line = rst[a:matcher.pattern.line]
+      let column = rst[a:matcher.pattern.column]
+      let message = rst[a:matcher.pattern.message]
```
thawk: 1. Suggest using temp variable to hold matcher.pattern.*, so it may save some time for large output (Maybe tens of thousand lines for a large project).
2. Suggest using ``get()`` to access ``matcher.pattern`` to deal with some cases like lacking ``column`` field
3. Suggest using ``get()`` to access ``rst`` or judge whether the corrsponding fields exists. Because not all error lines contains all four file/line/column/message fields

wsdjeg: Ok. updated.

PR #4175, opened 2021-04-06
Add 3 new keybindings begining with <Leader> to meet the documentation.

xylinkun:
### PR Prelude

Users may use these 3 key bindings in a normal model probably.

### DIFF
--- a/autoload/SpaceVim/default.vim
+++ b/autoload/SpaceVim/default.vim
@@ -156,6 +156,7 @@ endfunction
 function! SpaceVim#default#keyBindings() abort
   " yank and paste
   if has('unnamedplus')
+    nnoremap <Leader>y "+y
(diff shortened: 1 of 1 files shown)


wsdjeg:
leader y is only for visual mode.

xylinkun:
> leader y is only for visual mode.

Why?
Vimer used to copy everything in normal mode.

wsdjeg:
Because `"+y` in normal mode just do nothing, and it only switch into operate mode.

xylinkun:
> Because `"+y` in normal mode just do nothing, and it only switch into operate mode.

Then, "d, c, y" do nothing either but switch into operate mode. Do you agree?

wsdjeg:
yeah, but these key binding is not used for normal mode.

PR #4207, opened 2021-04-13
Fix P key binding in defx

wsdjeg:


mistical2008 (approved):
Works for case with copy by `yY`.  Cases with copy by `c` and moving by `m` don't. Just copies to clipboard without an opportunity to paste from.
Anyways, this is step forward.

PR #4217, opened 2021-04-18
Improve coc support

### LINKED ISSUE #2564: How to use coc.nvim in SpaceVim
How can i add coc.nvim support for autocompletion/lsp to SpaceVim. Tried many times but failed each time. I use it mostluy React/ReactNative/Nodejs. Thanks for great project :)

wsdjeg:
close https://github.com/SpaceVim/SpaceVim/issues/2564


Gabirel:
Ignoring `coc-settings.json` only in SpaceVim is not a good solution for users. 
User needs to save `coc-settings.json` somewhere else since this json file stores the user's config.

IMHO, it is a bad practice to add `coc-settings.json` in SpaceVim's gitignore since users may forget to copy this file when backuping. We could let user change their own `coc_config_home`.

PR #4284, opened 2021-06-12
Add split flygrep horizontally across vertical split

jsecchiero:
Hi There,

This minuscule PR allow flygrep to split horizontally across vertical windows so can be consistent with the other plugin/panels and also because the flygrep output requires a lot of horizontal space


wsdjeg:
thanks, merged.

PR #4343, opened 2021-07-23
Fix coffeescript api sys dependency

dukz:
Adding the lang#coffeescript layer throws an error.

Last commit broke this layer as the `s:SYS` is called before the `#api#import` code.

Swapped the two lines to fix the problem.


wsdjeg:
thanks.

PR #4388, opened 2021-08-12
Add: call into core layer to copy present line into clipboard.

JavierAcevedoC:
## Core Layer, Buffer Mapping
Add call to copy the present line from buffer into clipboard.


wsdjeg:
thanks for contributing, I think `noautocmd normal! "+yy` is better,  `normal! 0v$"+y'` will not include the space at the begin and end of current line. also autocmd should not be executed. doc also should be updated.
--------------
wsdjeg

PR #4391, opened 2021-08-13
Add lua file api

wsdjeg:


lua/spacevim.lua
```diff
@@ -23,13 +23,31 @@ end
...
+
 function M.call(funcname, ...)
     if vim.call ~= nil then
         return vim.call(funcname, ...)
     else
         if vim.api ~= nil then
             return vim.api.nvim_call_function(funcname, {...})
         else
+            -- call not call vim script function in lua
+            vim.command('let g:lua_rst = ' .. funcname .. '(' .. build_argv({...}) .. ')')
```
rafamadriz, line 49: As an alternative, you can declare global variables with `vim.g` see `:help vim.g`. So you could use:

```lua
vim.g.lua_rst = -- your value here
``` 
instead of using the `vim.command`

Not sure if there's any actual benefit from that, but since you're rewriting in lua , it's would be better to do things the lua way I suppose.

wsdjeg:
thanks, will try to do it.

PR #4532, opened 2021-11-29
Fix: Don't try to set guifont when using VimR

hakamadare:
### PR Prelude

The fix in #3457 is no longer necessary.  VimR no longer throws an error
message in response to the `guifont` command; however, VimR has its own
mechanism for font configuration, managed through the native GUI and
through some commands which can only be run from
`.config/nvim/ginit.vim`, neither of which can be invoked from this
file.

I found an acceptable workaround is setting `guifont` in my before hook,
  and entirely disabling the default `guifont` setting in `default.vim`.


wsdjeg:
try catch should not be removed.

PR #4648, opened 2022-04-25
feat(lang#c): add `SPC l g` to go to definition

sisrnb:
### PR Prelude

It's better to provide a shortcut to go to definition after user has pressed `SPC l` while editing c files. This is because the menu of `SPC l` has `declaration` but doesn't have `definition`, and `implementation` doesn't work with clangd.


wsdjeg:
@sisrnb for many languages layer, we use `g d` to jump to definition. and this key binding is available in `lang#c` layer.

sisrnb:
The menu of `SPC l` has `declaration` but doesn't have `definition`, the user goes here may want to go definition but there is not such an entry.
And `implementation` doesn't work with clangd.
And many language layers have shortcut `SPC l g`.
Also I personally think some consistency should be made between language layers.

sisrnb:
I was confused when I'm new to SpaceVim, I was wondering why `SPC l i` can not go to definition.

wsdjeg:
Maybe we should unify all these key bindings for different languages. maybe `g` is better for  `+Generate` group key bindings.

in development page: https://spacevim.org/development/#language-specified-key-bindings

sisrnb:
`lang#c` seems lack of maintenance, `SPC l i` (go to implementation) has no action c files.
I'm using clangd, I haven't tested any other c language server.

wsdjeg:
yeah, currently I am too busy, we also want to looking for contributors for each language layer. I will update the `lang#c` layer and doc.

PR #4655, opened 2022-04-29
feat(vim): support installing as a Vim package

### LINKED ISSUE #4653: Installing SpaceVim as a Vim 8 package

I configure my system using the [Nix package manager](https://nixos.org/). In Nix, there is a way to install Vim/Neovim onto the system. There's also support for configuring it with a custom .vimrc and custom Vim packages (those that can live in the `pack` directory in Vim 8). I'd like to leverage these building blocks to install SpaceVim.

I would like to be able to drop the SpaceVim repository underneath the `pack` directory of my Vim configuration to install it. I can then write my own install script (using Nix conventions) to do so for whichever version or variant of Vim I like.

**Describe alternatives you've considered**

There is [prior work](https://github.com/SpaceVim/SpaceVim/issues/3646) to create an installer for SpaceVim in Nix.

I believe that because SpaceVim _replaces_ the .vimrc/init.vim, it's bundled as a separate standalone application that wraps and replaces Vim (and not yet Neovim). The goal from this issue is more composable and a better fit for the Nix ecosystem.

**Additional context**

My goal ultimately is:

1.  Install SpaceVim as a standard Vim package.
2. Turn off plugin managers and auto-update so that my package manager configuration dictates which versions are in use.

I am consciously giving up convenience and getting changes immediately to get more flexibility in bundling other plugins and a more reproducible system.

Thank you for your time and consideration! I'm willing to work on this with some guidance if the goal is acceptable.

tjni:
### PR Prelude

This should fix #4653.

I have tried to make this change simple and small so that maintenance of it will be easy. I will be working on the other pieces I need in the upstream package manager to test this end-to-end, but this can be reviewed as-is.

Because this supports an advanced use case, I didn't see a clear place where documentation of this belongs. Please let me know if you would like me to write something about this and, if so, where.

### DIFF
--- a/main.vim
+++ b/main.vim
@@ -6,6 +6,22 @@
 " License: GPLv3
 "=============================================================================
 
+" Ensures that SpaceVim only initializes once.
+"
+" SpaceVim can be installed as a Vim package, in which case the entry point is
+" inside the 'plugin' directory and sources this script.
+"
+" SpaceVim can also be installed using the provided Vim entry point files (e.g.
+" vimrc, init.vim), which will source this script. However, the file inside of
+" the 'plugin' directory will still be executed and would run this twice if we
+" don't have this gate.
+"
+if exists('g:_spacevim_init')
+  finish
+else
+  let g:_spacevim_init = 1
+endif
+
 " set default encoding to utf-8
 " Let Vim use utf-8 internally, because many scripts require this
 set encoding=utf-8
--- a/plugin/SpaceVim.vim
+++ b/plugin/SpaceVim.vim
@@ -1,0 +1,9 @@
+"=============================================================================
+" SpaceVim.vim --- Entry point for SpaceVim installed as a plugin or package
+" Copyright (c) 2016-2022 Wang Shidong & Contributors
+" Author: Wang Shidong < wsdjeg@outlook.com >
+" URL: https://spacevim.org
+" License: GPLv3
+"=============================================================================
+
+execute 'source' fnamemodify(expand('<sfile>'), ':h').'/../main.vim'


tjni:
I tested this the whole way through my package manager, and it works!

For context, the way it will work in Nix is:

1. Nix provides Vim/Neovim packages that take plugins as a parameter, which will be installed as packages.
2. Vim plugins are packaged automatically by a script that takes in Git repo information.
3. Patches and customization (such as SpaceVim's TOML file), can be configured on top of this (via custom code).

Please let me know if you have any questions / concerns about this addition.

wsdjeg:
no, spacevim is not a plugin, so can not use it as plugin. it is based on the logic how vim load vim script file.

tjni:
> no, spacevim is not a plugin, so can not use it as plugin. it is based on the logic how vim load vim script file.

I don't see why it cannot be loaded as a plugin. In `init.vm`, for instance, all it does is source `main.vim`, which can happen also from a plugin entry point. Can you help me understand?

wsdjeg:
Read `:h startup`, you will find what is the difference between plugins and vimrc.

tjni:
Thank you for the reference. I read through it. Let me first say that I think it makes sense for vimrc to be how SpaceVim is installed by most people. It ensures that SpaceVim is the first code that initializes so that it can take control of the rest of plugin management, and also ensures all of the user's customization moves over to the TOML file.

Still, I don't see why SpaceVim cannot be loaded up as a plugin. I recognize there are some differences between `init.vm`, `ginit.vm`, and `vimrc`, which this patch currently does not capture, but nothing that SpaceVim needs to do on initialization that cannot happen from a plugin.

I don't want to change how SpaceVim is used by current users. I only want to provide this additional method of installation, because I think it composes better with the Nix approach of customizing Vim and Vim package installations.

If you disagree, I understand. I'll maintain a separate patch with this for my purposes. Thank you!

wsdjeg:
Sorry, I can not merge this PR, because this will cause many errors. For example, the plugin manager will not work. And the autocmd based on vim startup will not work. As I know both dein and vim-plugin do not support using as a plugin, all of them need to be used in vimrc.

tjni:
I have not run into those errors yet, so I'll live with the patch and see what happens. In the meantime, I'll close this PR.

Issue #4714, opened 2022-09-30
Getting error: Unable to create directory "/home/sonul/.cache/SpaceVim/swap" for swap file, recovery impossible: permission denied


sonulohani:
![image](https://user-images.githubusercontent.com/2958681/193182396-96ef4ab8-a0e0-4276-8a79-688b3ad3c65a.png)

jscd:
No idea why the cache dir is being made with the wrong permissions, but you can fix by adding read, write, and execute perms.
```sh
chmod -R u+rwx ~/.cache/SpaceVim
```

wsdjeg:
The directory is made by https://github.com/SpaceVim/SpaceVim/blob/master/autoload/SpaceVim/default.vim#L88-L102

ElonSatoshi:
> No idea why the cache dir is being made with the wrong permissions, but you can fix by adding read, write, and execute perms.
> 
> ```shell
> chmod -R u+rwx ~/.cache/SpaceVim
> ```

Are the execute perms necessary?

> The directory is made by https://github.com/SpaceVim/SpaceVim/blob/master/autoload/SpaceVim/default.vim#L88-L102

That's weird, I see 0700 on those, shouldn't that give the directory user read write and execute? 

This is what I'm getting when I install spacevim:

```
> ll -d ~/.cache/SpaceVim                                                                              23:53:31
d-w-r-xr-T 1 user user 0 Sep 30 23:53 /home/user/.cache/SpaceVim/
```



> No idea why the cache dir is being made with the wrong permissions, but you can fix by adding read, write, and execute perms.
> 
> ```shell
> chmod -R u+rwx ~/.cache/SpaceVim
> ```

I had to run this once, open and exit nvim, and run this once more in order for it to work.

ElonSatoshi:
> The directory is made by https://github.com/SpaceVim/SpaceVim/blob/master/autoload/SpaceVim/default.vim#L88-L102

That's weird, shouldn't that give the directory user read write and execute? 

This is what I'm getting when I install spacevim:

```
> ll -d ~/.cache/SpaceVim                                                                              23:53:31
d-w-r-xr-T 1 user user 0 Sep 30 23:53 /home/user/.cache/SpaceVim/
```

the-moisrex:
[...]

PR #4718, opened 2022-10-05
fix for lua removing leading zeros (#4714)

the-moisrex:
As I explained in [here](https://github.com/SpaceVim/SpaceVim/issues/4714#issuecomment-1268919712), leading zeros in lua are removed, thus resulting in a vim call of such:
```vim
:silent call mkdir("not/gonna/work", 'p', 700)
```
which results in the wrong permission.

Added quotes to make sure the leading zeros are kept.


wsdjeg:
@the-moisrex Thank you.

PR #4767, opened 2022-12-22
Add XDG support

### LINKED ISSUE #3517: XDG specification support
Please add support for XDG base directory specification:\
http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

I use to have all my configurations in `XDG_CONFIG_HOME` in order to make backups much simpler.\
In addition, having a clean `$HOME` is always a good idea.

It'd be nice if not only `~/.SpaceVim.d/` but also `~/.Spacevim/` were in `XDG_CONFIG_HOME`.\
Symlinks are not a smart solution for me. So I think the installer/updater script should move the locations.

Many world-renowned applications with a long history in Linx world address the problem overriding the default location using the environment variables (if set).

oliv3r:
### PR Prelude

XDG support is growing and more desireable, as has been discussed in #3517. Both vim and spacevim can work fine with the spec. This merge request offers (optional) support for XDG directory support.

Closes #3517

### DIFF
--- a/docs/install.sh
+++ b/docs/install.sh
@@ -89,6 +89,10 @@ Version='2.1.0-dev'
 System="$(uname -s)"
 # }}}
 
+XDGSpaceDir="${XDG_CONFIG_HOME:-${HOME}/.}${XDG_CONFIG_HOME:+/}SpaceVim"
+XDGvimDir="${XDG_CONFIG_HOME:-${HOME}/.}${XDG_CONFIG_HOME:+/}vim"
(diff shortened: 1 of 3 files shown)


docs/install.sh
```diff
@@ -346,6 +353,8 @@ download_font () {
 
 # install_fonts {{{
 install_fonts () {
+    need_cmd 'mkfontdir'
+    need_cmd 'mkfontscale'
```
mert574, line 357: these are not macos compatible

PR #4846, opened 2023-04-22
Feature: support`Telescope ultisnips` command if snippet engine is ultisnips

ChasingZenith:
Add more feature to the Telescope layer. Use `<Leader> f s` to `Fuzzy find Ultisnips snippets` if `snippets_engine="ultisnips"`.
Use [fhill2/telescope-ultisnips.nvim](https://github.com/fhill2/telescope-ultisnips.nvim).

The mapping is added to the doc.

Not 100% sure whether I can `add(plugins, ['fhill2/telescope-ultisnips.nvim', { 'merged' : 0}])`, a not bundled package.


wsdjeg:
merged, thanks, it is ok to use upstream plugin, and if this plugin is stable, we cal also use bundle feature.

PR #4850, opened 2023-04-25
fix: set default font correctly on macOS

hakamadare:
### PR Prelude

the problem here is that in Lua, `0` is truthy (see
https://www.lua.org/pil/2.2.html).  as a result, in
`lua/spacevim/default.lua`:

```lua
if SYSTEM.isWindows then
  ...
```

was always triggering, even when `SYSTEM.isWindows = 0`, which resulted
in an invalid default value for `guifont` on macOS.

a better fix would probably be to refactor `SYSTEM.isWindows`,
`SYSTEM.isLinux`, and `SYSTEM.isOSX` such that they are booleans rather
than numbers, i'll consider tackling that later, but this patch fixes
the immediate problem.

### DIFF
--- a/lua/spacevim/api/system.lua
+++ b/lua/spacevim/api/system.lua
@@ -27,7 +27,7 @@ end
 
 local M = {}
 
-if has('win16') ==1 or has('win32') == 1 or has('win64') == 1 then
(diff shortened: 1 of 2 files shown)


wsdjeg:
merged, thanks.

Issue #4836, opened 2023-04-13
Cannot create new file with neovim


dqnk:
## Expected behavior, english is required
On command `$ nvim test.c` I would expect nvim to open with Spacevim configuration and create a new file buffer (test.c does not exist currently). Instead Spacevim opens as if no file name was specified (startify menu).

## The reproduce ways from Vim starting (Required!)
Open Spacevim using `$ nvim test.c`.

When using nvim to create a new file, this line appears in :SPDebugInfo: `[ SpaceVim ] [00:00:00:449] [ Info  ] Startup with no argv, current dir is used: ~`
But with just Vim, this line appears instead: `[ SpaceVim ] [12:07:14] [   0.434] [  Info ] Startup with argv: 2`

If I just delete all Spacevim config, then both (barebone) nvim and vim create the file as expected.

This only happens with nvim version >= 0.9, with version 0.8.3 it works fine.

## Output of the `:SPDebugInfo!`
<details><summary> SpaceVim debug information </summary>

### SpaceVim options :

```toml
  auto_disable_touchpad = 1
  autocomplete_method = 'deoplete'
  autocomplete_parens = 1
  bootstrap_after = 'myspacevim#after'
  bootstrap_before = 'myspacevim#before'
  buffer_index_type = 4
  checkinstall = 1
  code_runner_focus = 0
  colorscheme = 'molokai'
  colorscheme_bg = 'dark'
  colorscheme_default = 'desert'
  commandline_prompt = '>'
  custom_color_palette = []
  custom_plugins = [['tbodt/deoplete-tabnine', {'build': './install.sh'}]]
  data_dir = '/home/dqnk/.cache/'
  debug_level = 1
  default_custom_leader = '<Space>'
  default_indent = 8
  disabled_plugins = []
  enable_bepo_layout = 0
  enable_cursorcolumn = 0
  enable_cursorline = 1
  enable_debug = 0
  enable_googlesuggest = 0
  enable_guicolors = 1
  enable_javacomplete2_py = 0
  enable_key_frequency = 0
  enable_language_specific_leader = 1
  enable_list_mode = 0
  enable_neocomplcache = 0
  enable_os_fileformat_icon = 0
  enable_powerline_fonts = 1
  enable_projects_cache = 1
  enable_statusline_bfpath = 0
  enable_statusline_mode = 0
  enable_statusline_tag = 1
  enable_tabline_filetype_icon = 1
  enable_tabline_ft_icon = 0
[...]

PR #4856, opened 2023-05-07
Fix: neovim not being able to create a new file #4836

dqnk:
### PR Prelude

Since Neovim version 0.9.0, SpaceVim could not create a new file with command `$ nvim new_file.ext` (#4836). This happened because argv list was never empty, but was treated as empty even though it had some arguments, because the `--embed` argument always appeared in the list. With this PR creating files works as intended.

Since I am new to vim language, it would be a good idea if someone checked that the string comparison is done right (should I be using a builtin function? Should I have expanded v:argv[-1] with `expand(v:argv[-1])`)? Do lowercase/uppercase matter?). Are there any other arguments that could appear apart from `--embed`?

It's just 2 short lines of code, so hopefully it can be merged soon.


wsdjeg:
Thanks :)

PR #4900, opened 2023-08-05
[FIX] Typescript LSP symbol rename functionality

dqnk:
### PR Prelude

This is a small fix for TS LSP.

Typescript LSP functionality of renaming symbols did not load correctly, because the if statement did not check if tsserver is an enabled client. This is an issue, because  the nvim-ts plugin does not handle a lot of cases such as shadowed variables and therefore often creates a mess. I have also reordered the if statements a little to give the user as much functionality from the LSP and nvim-typescript plugin as possible, but given LSP a higher priority, because it implements functionalities better.


wsdjeg:
merged, thanks.

PR #4941, opened 2024-12-07
fix(buffer): fix `vim#buffer` api

wsdjeg:


wsdjeg:
No need to support old vim!

